PR #14, opened 2016-12-20
some updates to resources.js

thomas-maschler:
- landsat annual layer
- glob cover
- some language changes


thomas-maschler:
not sure if commit 58ca8e33f5421c95c9f3e9a81b6794b61e249d8e is necessary or not. I need the lines to successfully test the changes, however, I don't see this lines in the develop and master branch. Are they added during the build?

thomas-maschler:
also removed the localhost token. it works with any machine and shouldn't be public

SwampGuzzler:
Per your commit 58ca8e33f5421c95c9f3e9a81b6794b61e249d8e change, this shouldn't happen in the es6 code in the 'src' directory, as it occurs automatically during any 'build' or 'dist' task when we're developing or building the app. 

Removal of the localhost token is fine!

thomas-maschler:
@SwampGuzzler ok, reverted the commit and updated the Portuguese translation.

PR #35, opened 2017-06-28
Add georgian language

pkmoran:
@SwampGuzzler


pkmoran:
@tgirgin23 I made the changes you requested, can you take another look when you get a chance?

Also, looks like we have conflicts again

tgirgin23:
@pkmoran  I can see some more changes that needs to be done, for example on line L221 https://github.com/wri/gfw-mapbuilder/blob/74abc9966e1ab4d8d722866704eb4e3adfdadc37/src/js/languages.js#L221 I believe we are missing the Georgian language once again.

For the languages.js, it seems it is missing in a few spots. Try searching for "ZH" through the file, and if you don't see "KA" for Georgian language right below it, it is probably a missing property.

resources.js looks good!

pkmoran:
@tgirgin23 fixed languages.js, should be ok now

PR #37, opened 2017-07-05
Using our URL's appid when loading the app as a library

SwampGuzzler:


pkmoran:
@SwampGuzzler ran through this with me. LGTM

PR #38, opened 2017-07-06
Custom legend

tgirgin23:


src/js/components/LegendPanel/LegendPanel.js
```diff
@@ -67,36 +51,114 @@ export default class LegendPanel extends Component {
...
+      return 0;
+    }
+    legendInfos.sort(compare);
+
     return legendInfos;
   }
 
+  createLegend = (layerDiv, index) => {
+    let childComponent;
+    const {map, language, settings} = this.context;
```
SwampGuzzler: So, if we are grabbing all of these heavy things from context just to pass them down to this same single childComponent, we would be better served grabbing them from context one step lower, in that childComponent, and avoiding all of these unnecessary props.

src/js/components/LegendPanel/LayerLegend.js
```diff
@@ -0,0 +1,70 @@
...
+    if(utils.getObject(layerGroups.GROUP_LC.layers, 'id', this.props.layerId)) {
+      layerConf = utils.getObject(layerGroups.GROUP_LC.layers, 'id', this.props.layerId);
+    } else if(utils.getObject(layerGroups.GROUP_LCD.layers, 'id', this.props.layerId)) {
+      layerConf = utils.getObject(layerGroups.GROUP_LCD.layers, 'id', this.props.layerId);
+    }
+
+    if(this.state.visible === false) {
+      bool = 'hidden';
+    } else {
+      bool = '';
```
SwampGuzzler: No es necesario

PR #41, opened 2017-07-13
Carto implementation

tgirgin23:


src/js/layers/CartoLayerBase.js
```diff
@@ -0,0 +1,358 @@
...
+          //project geometry to web mercator if needed
+
+          if (graphic.geometry.spatialReference.wkid === 4326){
+            graphic.setGeometry(
+              webMercatorUtils.geographicToWebMercator(graphic.geometry)
+            );
+          }
+
+          // Set the symbolDictionary depending on the geometry type
+          if (!!this.symbolDictionary) {
```
SwampGuzzler: Ooh I'm not sure I like the double negative...what exactly are we checking for here?

tgirgin23: This part of the code has been copied directly from odoenet's carto implementation

SwampGuzzler: Ok we can talk when I get back!

PR #52, opened 2017-07-28
Add modis fires

pkmoran:
@SwampGuzzler


src/js/actions/LayerActions.js
```diff
@@ -30,7 +30,11 @@ class LayerActions {
     return info;
   }
 
-  changeFiresTimeline (selectedIndex) {
+  changeViirsFiresTimeline (selectedIndex) {
```
SwampGuzzler: TODO: Decide whether we still need these timelines now that we have the start and end date pickers for each Fires layer. @csphang

src/js/components/LayerPanel/FiresControls.js
```diff
@@ -1,50 +1,139 @@
...
+  }
+
+  componentDidMount() {
+    //- Load the pickers css
+    let base = window._app.base ? window._app.base + '/' : '';
+    if (base && base[base.length - 1] === '/' && base[base.length - 2] === '/') {
+      base = base.substring(0, base.length - 1);
+    }
+
+    loadCSS(base + assetUrls.pickadateCSS);
```
SwampGuzzler: Let's do this just the once (rather than here and in our GLADControls.js ), higher up in the component tree, probably just where we load the rest of our libraries css.

PR #54, opened 2017-07-31
implement range slider for tc loss and visualizations over time

pkmoran:


src/js/components/LayerPanel/LossControls.js
```diff
@@ -51,7 +51,6 @@ export default class LossControls extends Component {
         $('#loss-slider').ionRangeSlider({
           type: 'double',
           values: lossOptions.map(option => option.label),
-          min_interval: 1,
```
SwampGuzzler: Let's confirm with Christina: I think this is correct and the user should be allowed to only see data from one year:  our TreeCoverLossLayer is inclusive-inclusive on its date def query!

PR #56, opened 2017-08-02
674: clear all button resets the layer panel state

pkmoran:
Do not merge until the following have happened:

- [x] #55 is merged
- [x] This PR is updated to reset the Terra I calendar picker in the layer panel


src/js/components/LayerPanel/TerraIControls.js
```diff
@@ -84,10 +78,20 @@ export default class TerraIControls extends Component {
...
   }
 
+  componentDidUpdate(prevProps) {
+    //ensure the startDate is an empty object and not a Date Object
+    if ((prevProps.startDate !== this.props.startDate && this.props.startDate.constructor === Object)
+      && prevProps.endDate !== this.props.endDate && this.props.endDate.constructor === Object) {
+      this.fromPicker.set('select', this.state.min);
+      this.toPicker.set('select', this.state.max);
+    }
+  }
```
pkmoran: Terra I works a little different than GLAD here. Since the min and max dates are set when the layer is initialized (instead of in the initial state), we can't just reset to the initial state when the clear button is clicked. We need a way to tell the component that the clear button has been pressed, so I chose to pass empty objects for `startDate` and `endDate`. Then we need to check if the component received an empty object, and update the calendar (and layer def) accordingly.

src/js/components/LayerPanel/TerraIControls.js
```diff
@@ -96,16 +100,19 @@ export default class TerraIControls extends Component {
...
       }
     }
   };
 
-  updateDateRange = () => {
-    const {layer, startDate, endDate} = this.props;
+  updateDateRange = (startDate, endDate) => {
+    const {layer} = this.props;
+    startDate = startDate.constructor === Date ? startDate : this.state.min;
+    endDate = endDate.constructor === Date ? endDate : this.state.max;
```
pkmoran: same thing happening here. Since this is called once for each time the start or end is set separately, we will always have one date and one empty object when the clear button is pressed. To fix that we just do a check and return either the param passed or the min/max in state.

PR #75, opened 2017-09-26
re #332: add subtitle questions for indicators sections onclick

pkmoran:


pkmoran:
Spoke with @SwampGuzzler - OK to merge

PR #78, opened 2017-09-26
re#330 update legend order to respect layer panel order

pkmoran:


pkmoran:
This change doesn't work for the `Land Cover Dynamics` group

PR #94, opened 2017-10-20
Add support for circles and lines in legend. Fix legend ordering

pkmoran:
@SwampGuzzler pretty confusing diff here, let me know if you have questions


SwampGuzzler:
This code is all working great. Howmever, while our legend and our layer-panel are showing the correct layer order (from the webmap), our actual map-layers are ordered incorrectly! Let's peep our 
map.reorderLayer(map.getLayer(layer.id), layer.order);
in MapActions to see if this is still necessito

PR #125, opened 2017-12-13
Indigenous Lands Report via dropdown

SwampGuzzler:


src/js/components/AnalysisPanel/RedirectChart.js
```diff
@@ -0,0 +1,22 @@
+import React, { Component } from 'react';
+
+
+export default class RedirectChart extends Component {
+
+  componentDidMount() {
+    const openWindow = window.open('/country-profiles/');
```
pkmoran: note for the future: maybe we can make this a prop so we can reuse this component if needed?

pkmoran:
:shipit:

PR #128, opened 2017-12-18
fix tree cover loss not showing in report

pkmoran:
* [Build to test this here](http://alpha.blueraster.io/gfw-mapbuilder/712-tcloss-report/)


src/js/layers/TreeCoverLossLayer.js
```diff
@@ -37,8 +37,8 @@ export default declare('TreeCoverLossLayer', [TileCanvasLayer], {
   },
 
   // Filter Data Method
-  filter: function(data) {
-    const z = brApp.map.getZoom();
```
pkmoran: The report didn't know what `brApp` was, so I'm just passing the zoom level to the filter function now

PR #130, opened 2017-12-19
Wacky wordpress fix

pkmoran:


SwampGuzzler:
Although fantastic, this PR is unnecessary right now! We are keeping both of these branches for posterity since once is Wordpress-centric and the other doesn't cater to pure evil.

PR #147, opened 2018-01-16
Bring in custom groups

pkmoran:


SwampGuzzler:
This looks pretty good so far, but we're going to leave it open until Christina has vetted it across different webmap/appId combinations.

PR #185, opened 2018-07-11
add s3 plugin for automatic deploys on production build

pkmoran:


SwampGuzzler:
Spectaculous. Once @csphang gives us the go-ahead, we can merge this in. We can also wait until we have our deploy script set up for a more streamlined process or move that to a different PR.

SwampGuzzler:
Looking awesome, wait to put this into action next week!

pkmoran:
this is no longer only for the library build. This is now an all in one deploy script. It will update gh-pages for the production build. It will generate a build and put it on wri's s3. And it will generate a library build and put it on wri's library s3.

PR #202, opened 2018-08-31
Develop

peterxiao1234:


SwampGuzzler:
We will be doing this, but only after our old `...config.json` files have been updated or removed and with the publication of a major or minor release.

Issue #205, opened 2018-10-02
Confirm report fix is deployed for Monday's launch


csphang:
- [x] Report fix (cmr.forest-atlas.org/map 
http://eth.doesntexist.org/Atlas)

**Priorities for the report:**
1. Formatting - add extra margin around analyses
2. Order - analyses change depending on which one comes back first.
3. Refresh - ensure report re-loads upon browser refresh **(DONE)**
4. Sharing - permanent link so report re-hydrates when opening in new browser or sharing with someone

SwampGuzzler:
https://beta.blueraster.io/mapbuilder/eth-exists-report/external.html

SwampGuzzler:
https://beta.blueraster.io/mapbuilder/cameroon-report/external.html

thomas-maschler:
@SwampGuzzler The ethiopia map uses custom analysis (none of the standard GFW analysis are activated). Please make sure the report shows the same analysis as those listed in the config file.

SwampGuzzler:
https://beta.blueraster.io/mapbuilder/external-library-test/external.html

SwampGuzzler:
https://beta.blueraster.io/mapbuilder/eth-report-analysis/external.html

PR #213, opened 2018-10-03
Library app base

SwampGuzzler:


SwampGuzzler:
@jenboyd See the builds that I was worried about succeeding on this issue: #205

Issue #256, opened 2018-10-31
Layer order for groups changes when you translate the applications


richardbarad:
If you open an application in the primary language and then switch to the secondary language the order of layers which are not coming from the webmap get reversed. 

For an example, see the screenshoot below taken from the Cameroon atlas:

https://cmr.forest-atlas.org/map

![2018-10-31 12_26_31-window](https://user-images.githubusercontent.com/40440743/47803573-2f21f980-dd09-11e8-8b53-c527930fa24d.png)

![2018-10-31 12_26_31_2](https://user-images.githubusercontent.com/40440743/47803670-61335b80-dd09-11e8-86e6-f1c613d1d401.png)

SwampGuzzler:
http://beta.blueraster.io/mapbuilder/256-layer-order-translate/external.html

csphang:
@SwampGuzzler This is fixed for the Land Cover Change Dynamics layer menu group, but it's now occurring in the "Land Use" layer group. 

![2018-11-06_14-23-54](https://user-images.githubusercontent.com/13664037/48088578-63e1f500-e1d0-11e8-9caf-fee655c02048.png)

![2018-11-06_14-24-51](https://user-images.githubusercontent.com/13664037/48088586-66dce580-e1d0-11e8-9850-6829f41218c2.png)

SwampGuzzler:
@csphang  https://beta.blueraster.io/mapbuilder/256-layer-order-groups/external.html

csphang:
Tested and confirmed this is resolved. Ready for final review at: https://beta.blueraster.io/mapbuilder/256-layer-order-groups/external.html

jenboyd:
Released with version 1.3.1

PR #259, opened 2018-11-05
Ensuring we only re-order our layerPanel layers on 1st map

SwampGuzzler:
Addresses issue #256 of course!


SwampGuzzler:
http://beta.blueraster.io/mapbuilder/256-layer-order-translate/external.html

PR #277, opened 2018-11-26
In Progress: Satellite Imagery

jenboyd:


SwampGuzzler:
@jenboyd Looking awesome. One question: do we still need an `AppStore` listener on `toggleImageryActive`?

src/js/utils/request.js
```diff
@@ -204,9 +207,148 @@ const request = {
...
+    });
 
+    // Create request body that will be used for both the recent-tiles/tiles
+    // request and the recent-tiles/thumbs request
+    const content = {
+      bands: params.bands,
+      source_data: sourceData,
+    };
+
+    const postTiles = (count = 0) => {
```
SwampGuzzler: @jenboyd Can we hoist `postTiles` and `postThumbs` up to the same independent level as their parent functions so that they are not defined inside of a `request` helper function?

PR #278, opened 2018-11-29
263 center and zoom

SwampGuzzler:
Initial Center and Zoom of the Map Extent can now be set in our constructor params (`resources` or `library builds`)


SwampGuzzler:
We can test this [here](https://beta.blueraster.io/mapbuilder/263-initial-location/external.html) to make sure we start off in **Seattle**!

Issue #317, opened 2019-01-22
Layer panel - allow users to select/search from a list of unique values for a specified feature service field


csphang:
Support feature layers and dynamic layers.

User specifies the service URL and the field that should be queried in the parameter. Labels use the field alias for display.

### Example
Endpoint URL: https://gis-gfw.wri.org/arcgis/rest/services/partner_requests/greenpeace_en/MapServer/0
Field: po_grup
Field alias: Group

See example from ZSL SPOTT:
![2019-01-22_16-24-35](https://user-images.githubusercontent.com/13664037/51566417-7a5c2c00-1e62-11e9-8aa4-14c6e324f7b4.jpg)

jenboyd:
@csphang link to test: http://alpha.blueraster.io/gfw-mapbuilder/317-layer-filter-fields/

Layer names are `Palm Oil Concessions - feature layer` and `Palm Oil Concessions - dynamic layer`

The dynamic layer can accept a multiple sublayers, but all sublayers would need to have the same filter field.. is this what you are thinking?

csphang:
Continue to support dynamic layers, but will only filter attributes based on one field.

richardbarad:
Hi @csphang, @SwampGuzzler , and @jenboyd.

This looks great! I just shared it with our collague who manages the RSPO mapbuilder app and he is very excited about being able to filter his palm oil concessions by company? https://www.rspo.org/geo-rspo

I took a look at the configuration and had a few questions:

1) Could we make the "filter by group" text adjustable in the configuration? I could see instances where someone might want to "filter by type" or "filter by status" depending on what layer they are filtering.

2) Also, how hard would it be to support multiple languages here? Both, for the text label and for the field. One of the teams which is interested in this functionality is Georgia and thier application is available in multiple languages. 

Richard

jenboyd:
Link updated with comments above: http://alpha.blueraster.io/gfw-mapbuilder/317-layer-filter-fields/

jenboyd:
Example of config for both feature and dynamic layer types
```
 {
        order: 9,
        id: 'PALM_OIL',
        type: 'feature',
        url: 'https://gis-gfw.wri.org/arcgis/rest/services/partner_requests/greenpeace_en/MapServer/0',
        visible: false,
        label: {
          en: 'Palm Oil Concessions - feature layer',
          fr: 'Palm Oil Concessions - feature layer',
          es: 'Palm Oil Concessions - feature layer',
[...]

PR #325, opened 2019-02-05
Sprint 5 -  v1.3.4 Add layer filter field option to layer panel

jenboyd:
#317 

http://alpha.blueraster.io/gfw-mapbuilder/317-layer-filter-fields/

****NOTE: We will need to remove the layer config examples from resources & external.pug when we are finished testing since these aren't part of the default mapbuilder config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
   build:
     docker:
       # specify the version you desire here
-      - image: circleci/node:6.11-browsers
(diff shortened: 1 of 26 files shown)


.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   build:
     docker:
       # specify the version you desire here
-      - image: circleci/node:6.11-browsers
```
SwampGuzzler: This is awesome

Issue #318, opened 2019-01-22
Support custom basemaps


csphang:
Support additional basemaps! Several use cases:

1. User has a basemap in different language
2. User creates their own basemap and would want to see it
3. User selects an Esri Basemap not normally a part of MapBuilder (for example, topographic)

richardbarad:
Hi @csphang,

For India here is the basemap they would like to use: http://gfw.maps.arcgis.com/home/webmap/viewer.html?webmap=acb6f60ce2fe469c9a1aa92de084c264

Just wanted to make sure we are on the same page regarding our approach here. 

I think we could have mapbuilder set up such that mapbuilder defaults to whatever basemap is used in the webmap even if it is a custom one and the name used is the same name specified in the basemap? 
The main question I have is how do we approach getting a thumbnail for custom basemaps. Any ideas?

Also, if they want to have a 2nd custom basemap could they specify it in the configuration file? If so, would  it be better to point to an ArcGIS webmap congaing a basemap or point directly to tiles or web service?

Best,

Richard

csphang:
@richardbarad Want to confirm we're interpreting this correctly! Would you want **both** the topographic and the India Boundary layers as the basemap? Or just the India Boundary layer? In this India example, neither the Topographic nor the India boundary are in our current Basemap options. But what about the scenario where a user has a webmap with two basemap layers and one layer is already supported (for example, India Boundary and Imagery)? Would the app display the same way and include both layers as the basemap?

![2019-01-28_13-40-37](https://user-images.githubusercontent.com/13664037/51858856-b7fa0280-2303-11e9-8807-69b2f155ac6c.jpg)

richardbarad:
@csphang,

Let me check with the India to confirm but I think they would want to show both the topographic and the India boundary layers.

richardbarad:
Hi @csphang,

Confirming that the India team would like to use both layers as the basemap.

SwampGuzzler:
@richardbarad Check out a build addressing this [here!](http://alpha.blueraster.io/gfw-mapbuilder/318-custom-bm/external.html)

I would still like to see a webmap with other custom basemaps for more ways to test my fix, do you know of any other builds like this?

richardbarad:
Thanks @SwampGuzzler.

It looks like the White India layer is showing up on top of the other layers instead of as part of the basemap.

[...]

PR #352, opened 2019-03-18
318 custom basemaps

SwampGuzzler:
Addresses #318 ! Test-able [here](http://alpha.blueraster.io/gfw-mapbuilder/318-custom-bm/external.html)


SwampGuzzler:
UNWORTHY! (fails via our Cameroon set-up and probably others as well)

PR #408, opened 2019-06-25
Wcs desktop modules debugging

SwampGuzzler:
Fixes our [beta map's](https://measures.wcs.org/beta/map) `Species: Population Trend` analysis by adding in the correct `featureDataFieldsToPass` in the *analysisModule* and adding the module's `analysisId` to the API Call


SwampGuzzler:
@SwampGuzzler Can we update the [Analysis Docs file](https://github.com/wri/gfw-mapbuilder/blob/wcs-desktopModules-debugging/ANALYSIS.md) to show exactly how the WCS Analysis works with the `featureDataFieldsToPass` and `analysisId ` fields in this new commit?

Issue #395, opened 2019-06-04
Add a description for each chart


csphang:
Use metadata endpoint -- Richard to post the metadata

richardbarad:
I posted metadata to these widgets:

- Total Tree Cover Loss/Gain - `95c2c559-ca78-4b7a-b18b-7b2bca14ce83`
- Annual Tree Cover Loss - `e53e541c-92cd-4b00-9aa7-2c7bb36d4697`
- CO2 Emissions from biomass loss -`ac38fdbd-fdb1-4d8e-9109-674013fb51a2`
- Annual Tree Cover Loss in IFL - `2083a1bc-440d-43fe-8b50-ff9918a37c57`
- Annual Tree Cover Loss by Land Cover Class - `31f78466-fc0b-42f9-a7ae-bea8559740d8` 
- GLAD Alerts Per Month - `0734ba0a-3a6c-4388-aa4a-5871791b1d1f`
- Total GLAD Alerts - `16ff6282-8ceb-4055-938a-43726a62b205` 
- VIIRS Active Fires - `5d696f96-e6c7-4323-8bda-4c99cd6b0cb4`
- Land Cover Composition - `1b84364d-0efd-4d60-81ef-870f7d13ee7b`

It looks like there are several ways to pull the metadata:

This calls returns the widget + metadata in all languages:

http://production-api.globalforestwatch.org/widget/95c2c559-ca78-4b7a-b18b-7b2bca14ce83?includes=metadata

https://production-api.globalforestwatch.org/v1/dataset/a97036a8-5526-4b85-8ffb-9ee2cdd9b34a/widget/95c2c559-ca78-4b7a-b18b-7b2bca14ce83?includes=metadata

This call returns **just** the metadata without the widget and you can add a filter by language.

https://production-api.globalforestwatch.org/v1/dataset/a97036a8-5526-4b85-8ffb-9ee2cdd9b34a/widget/95c2c559-ca78-4b7a-b18b-7b2bca14ce83/metadata

https://production-api.globalforestwatch.org/v1/dataset/a97036a8-5526-4b85-8ffb-9ee2cdd9b34a/widget/95c2c559-ca78-4b7a-b18b-7b2bca14ce83/metadata?language=en

Unfortunately, this call does not seem to work so it looks you have to include the dataset ID in the query if you want to retrieve **just** the metadata.

[...]

PR #420, opened 2019-07-08
Retrieve chart description from API call

KaylaKremer:
https://github.com/wri/gfw-mapbuilder/issues/395


SwampGuzzler:
@KaylaKremer Don't make the `description` API call if we are Not in the report!

Issue #406, opened 2019-06-11
Fires layers - popup issue


csphang:
Pop-ups only work for fires from the past 24 hours. 

If you switch to another time period and click on a fire point which older than the past 24 hours the pop-up does not show up. I am attaching a video in which I switch the time period to the past week and once I do this some of the pop-ups stop working. 
[firepopups.zip](https://github.com/wri/gfw-mapbuilder/files/3273906/firepopups.zip)

KaylaKremer:
@SwampGuzzler Did you already fix this?

SwampGuzzler:
@KaylaKremer No! This is a new issue!

richardbarad:
@SwampGuzzler & @KaylaKremer,

I tested this in this build: http://alpha.blueraster.io/gfw-mapbuilder/406-fires-bug/, which I think is 

It looks like the pop-ups are working when you select past 24 hours or past 48 hours, but not when you select past 72 hours or past week. I see the same behavior on both VIIRS and MODIS. Please see attached video. 

[firesbug.zip](https://github.com/wri/gfw-mapbuilder/files/3526527/firesbug.zip)

KaylaKremer:
@richardbarad I think I haven't updated that build for a while, but I believe we solved this issue. We had to update the VIIRS 1 YR fire service (We use this service for anything beyond 72 hours). We had only updated the service on staging, so we just need to update it on production. I can provide an updated build once this is done.

KaylaKremer:
@richardbarad I've gone ahead and updated the build just now for you to test again and be sure. It's still using our 1 YR VIIRS service on staging, which we plan to switch out for the production one:

http://alpha.blueraster.io/gfw-mapbuilder/406-fires-bug/

richardbarad:
@KaylaKremer. I just checked this on MODIS and it looks there is still an issue. If I:

1. Turn on the MODIS Active Fires layer
2. Zoom in on the map to central DRC on an area with a large number of fires
3. Toggle from past 24 hours to 48 hours, and identify a point which was not present in the 24 hours layer
4. Click on that point
5. The point does not show any results in the attribute table

I noticed that it looks like the query is always being made on the `https://gis-gfw.wri.org/arcgis/rest/services/Fires/FIRMS_Global_MODIS_24hrs/MapServer/21` service even when you click on a point in the 48 hours service. See screenshot below and the attached video. 

[...]

PR #424, opened 2019-07-12
406 fires bug

kaylakremer:
#406

http://alpha.blueraster.io/gfw-mapbuilder/406-fires-bug/

### DIFF
--- a/src/external.pug
+++ b/src/external.pug
@@ -6,7 +6,7 @@ html
 	body.tundra
 		#root
 		div#share-modal.modal-wrapper.hidden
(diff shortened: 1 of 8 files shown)


SwampGuzzler:
@KaylaKremer This functionality looks good, but we can't merge this in until the **Production** ArcGIS Server has the new service update! Right now we are pointing to the `https://gfw-staging.wri.org/arcgis/rest/services/Fires/FIRMS_Global_VIIRS_1yr/MapServer` and we need to point it back to prod before we hit merge!

PR #426, opened 2019-07-15
Loader in library build to avoid svg flashes

SwampGuzzler:
http://alpha.blueraster.io/gfw-mapbuilder/smoother-loading/


SwampGuzzler:
Our latest commit and update sets up [smoother-loading-v2](http://alpha.blueraster.io/gfw-mapbuilder/smoother-loading-v2/), which is doing the same thing but keeping the `external.pug` file clean (which is important because this turns into the `index.html` in our library) and easy to manage from the user's perspective.

Issue #433, opened 2019-07-22
(Metadata): if null metadata, hide grey box


csphang:
![2019-07-22_17-19-28](https://user-images.githubusercontent.com/13664037/61666237-f8c84800-aca4-11e9-9981-3e3f23ca1d0d.png)

csphang:
Verified on Ethiopia consolidated build:
http://alpha.blueraster.io/gfw-mapbuilder/new-report-consolidated-builds/ethiopia/

![2019-08-20_14-49-19](https://user-images.githubusercontent.com/13664037/63375424-be140700-c359-11e9-9237-5d4c025a6a85.png)

PR #448, opened 2019-07-24
433 hide metadata if null

KaylaKremer:
#433 

http://alpha.blueraster.io/gfw-mapbuilder/433-hide-metadata-if-null/


KaylaKremer:
Accidental PR duplicate!

PR #455, opened 2019-07-26
Share modal dom removal

SwampGuzzler:
Our initial `index.pug` and (more importantly) our `external.pug` used to have this weird `div#share-modal` html next to our **root** dom node. Since there is nothing special about the Share Modal, I brought that into our app as a more generic modal and changed its visibility to rely on **alt** via the `ModalActions` and the `MapStore`.

I want the `external.pug`, which turns into the `index.html` that the client sees to be as clean as possible!

### DIFF
--- a/src/css/modal.styl
+++ b/src/css/modal.styl
@@ -90,8 +90,8 @@
       margin 5px 0
 
 //- Share Modal
-#share-modal
-  & .modal-content
+.share-modal-wrap
+  &.modal-content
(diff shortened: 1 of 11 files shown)


SwampGuzzler:
**HOLD** until we get the `new report build` in and tested!

SwampGuzzler:
@SwampGuzzler Great idea! But outdated ):

Issue #439, opened 2019-07-23
Provide updated design specifications for the report.


richardbarad:
Provide updated design specifications for new report without the grey space next to the map.

![2019-07-23_12-21-20](https://user-images.githubusercontent.com/13664037/61729077-7db66e80-ad44-11e9-92de-f9e8b856a00b.png)

richardbarad:
@csphang. Do you think we need to include the grey box which includes the text **"Area of Analysis feature analysis"** ? I think we could remove the grey box and do the following:

1. If it is a custom area show the text "Custom Feature" under Area of Analysis. 

2. If it is an area from the map display the name of the layer that the users clicked on. For example, if analyzed a mining concession. I would see the text "Mining Concession" above the table instead of "Feature Analysis".

I suggest we center the text above the table.

csphang:
![updated-report-header](https://user-images.githubusercontent.com/13664037/61743560-8d917b00-ad63-11e9-8ae3-da9628d1d818.png)

richardbarad:
Thanks @csphang.

Do you think we need to include the grey box as shown below? I thought we could center the text and make it larger to eliminate some of the empty space. 

![2019-07-24 05_51_30-Window](https://user-images.githubusercontent.com/40440743/61784435-b05d7700-add7-11e9-89ab-3ccd9742a336.png)


Also, since this was a feature of the map which the user clicked on I think should show the name of the map layer which contained the analyzed feature instead of **"custom feature"**. For example, it looks like the shape above is a Mining Concessions. Let me know if that is not clear!

KaylaKremer:
Hi @richardbarad ,

I can go ahead and make the style changes you want for this! I'll remove the gray box and center align and enlarge the text. Also, I believe how I have it set up now, it will grab the name of the map layer if there is one. But I can add this in if it's not.

richardbarad:
@KaylaKremer. Thanks! Please proceed.

PR #477, opened 2019-08-02
439 get map layer name

kaylakremer:
#439 

http://alpha.blueraster.io/gfw-mapbuilder/439-get-map-layer-name/

### DIFF
--- a/src/js/components/Map.js
+++ b/src/js/components/Map.js
@@ -160,6 +160,17 @@ export default class Map extends Component {
   storeDidUpdate = () => {
     this.setState(MapStore.getState());
   };
+  
+  getSelectedFeatureTitles = () => {
+    let selectedFeats;
+    const selectedFeatureTitlesArray = [];
+    if (brApp.map.infoWindow && brApp.map.infoWindow.getSelectedFeature()) {
+      selectedFeats = brApp.map.infoWindow.features;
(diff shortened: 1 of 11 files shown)


src/js/components/Map.js
```diff
@@ -174,6 +184,7 @@ export default class Map extends Component {
       response.map.graphics.clear();
       //- Attach events I need for the info window
       response.map.infoWindow.on('show, hide, set-features, selection-change', mapActions.infoWindowUpdated);
```
SwampGuzzler: @KaylaKremer Maybe listen for on(`hide`.. with a different callback that simply Clears Out the selectedFeatureTitles?

src/js/report/report.js
```diff
@@ -745,16 +745,14 @@ export default class Report extends Component {
   };
 
   render () {
-    const {analysisModules, mapForTable, paramsForTable} = this.state;
+    const {analysisModules, mapForTable, paramsForTable, selectedFeatureTitles} = this.state;
```
SwampGuzzler: the `selectedFeatTitles` will now be grabbed from the url!

src/js/utils/AppUtils.js
```diff
@@ -157,8 +157,7 @@ const utils = {
       viirsEndDate,
       modisStartDate,
       modisEndDate,
-      //This will hold the array of feature titles!
-      //selectedFeatureTitles
+      selectedFeatureTitle
```
SwampGuzzler: TITLES --> multiple!

Issue #482, opened 2019-08-05
Issues with Print formatting


richardbarad:
When using the new print functionality in report, the output is not formatted properly. Please see attached PDFs which use the new print functionality. One is generated in Chrome and the other in Edge.

The issues:

- The AOI is not completely visible in the map
-  There is overlap between the text descriptions, titles, and the widgets
- Can we structure this such that the widgets, title, description for a specific analysis are always  on the same page?

[Print_Chrome.pdf](https://github.com/wri/gfw-mapbuilder/files/3468167/Print_Chrome.pdf)
[Print_Edge.pdf](https://github.com/wri/gfw-mapbuilder/files/3468168/Print_Edge.pdf)

KaylaKremer:
@richardbarad I adjusted the CSS for the print style. Let me know what you think!
http://alpha.blueraster.io/gfw-mapbuilder/482-print-formatting-issue/

Also, I think the AOI issue was something @csphang noticed in general in our reports, not just in the print view. I'm not sure if a ticket was created or it got removed. But that is something we can discuss for tomorrow!

richardbarad:
@KaylaKremer. A few comments:

1. Can we hide the report titles when an analysis is turned off in the report? See what is currently looks like when I turn off the TCL Analysis.

2. It looks like the we also need to make sure the title for the first widget and the actual widget are always on the same page. Currently, the IFL title and the widget are on separate pages. See attached PDF.

Regarding the AOI issue, I had previously noticed this issue in some of the old builds as well, but I think it was fixed due to the report redesign by having the map take up the full width of the report.

From my testing, it looks like it is only present in the report now.

[Print_report.pdf](https://github.com/wri/gfw-mapbuilder/files/3468610/Print_report.pdf)

Richard

csphang:
@richardbarad the updates for this are testable in the consolidated builds. 

http://alpha.blueraster.io/gfw-mapbuilder/new-report-consolidated-builds/default/

![2019-08-21_11-38-41](https://user-images.githubusercontent.com/13664037/63447930-ee67ae00-c40a-11e9-829b-9d77bb56d6e3.png)
[report-from-consolidated-build.pdf](https://github.com/wri/gfw-mapbuilder/files/3526168/report-from-consolidated-build.pdf)

richardbarad:
@KaylaKremer. This looks good at the links above. I am still seeing the issue where the shape will sometimes not center in the printed report. Please see the attached PDF. 

[...]

PR #485, opened 2019-08-05
482 print formatting issue

KaylaKremer:
#482 

http://alpha.blueraster.io/gfw-mapbuilder/482-print-formatting-issue/

### DIFF
--- a/src/css/report.styl
+++ b/src/css/report.styl
@@ -626,15 +626,19 @@ body
 @media print
 
   html
+    min-height 100%
+    position relative
   body
-    // height 100%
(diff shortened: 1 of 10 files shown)


KaylaKremer:
@SwampGuzzler Build has been updated! The extra page and the TCL splitting into two pages should be resolved now. Each chart should be on their own page, centered. 

The only thing I'm still not quite able to figure out is the AOI issue. I find though that if you click print too quickly, it is off. But if you wait for at least some of the charts to load, then click print, it appears a lot better.

Issue #464, opened 2019-07-29
The toggled on layers do not display in the report


KaylaKremer:
The report doesn't retain the layers that have been turned on in the map.

Map:
<img width="870" alt="Screen Shot 2019-07-30 at 2 05 15 PM" src="https://user-images.githubusercontent.com/22095244/62153661-13b54080-b2d3-11e9-949e-e06cf9fae121.png">

Report:
<img width="650" alt="Screen Shot 2019-07-30 at 2 04 34 PM" src="https://user-images.githubusercontent.com/22095244/62153676-1d3ea880-b2d3-11e9-82a6-d2700b7f6027.png">

csphang:
Verified on consolidated builds. For example, testable on http://alpha.blueraster.io/gfw-mapbuilder/new-report-consolidated-builds/tierras-indigenas/

![2019-08-20_15-19-19](https://user-images.githubusercontent.com/13664037/63377336-ef8ed180-c35d-11e9-88b7-79f16d14acb8.png)

PR #491, opened 2019-08-08
464 toggled on layers not displaying

kaylakremer:
#464 

http://alpha.blueraster.io/gfw-mapbuilder/464-toggled-on-layers-not-displaying/


SwampGuzzler (approved):
Easy!

PR #516, opened 2019-08-26
Coordinates edit functionality

KaylaKremer:
WIP for the editing feature of the coordinates


KaylaKremer:
We decided at the last meeting with Richard to just display the latitude and longitude values of the coordinate when editing. I have a new branch to implement this functionality instead.

PR #524, opened 2019-09-05
Proper layerDefs for fires custom range filters

SwampGuzzler:
We had some different logic than `develop` in this `updateFiresLayerDefinitions` function and now I updated it so this will work for all dropdown selections as well as a **Custom Range**!

http://alpha.blueraster.io/gfw-mapbuilder/519-fires-custom-range/


kaylakremer (approved):
🔥 🔥 🔥

Issue #460, opened 2019-07-29
Support custom color theming


csphang:
Ensure this is configurable through the ArcGIS Online template as well!

richardbarad:
@richardbarad, to prepare list of what needs to be when you change the custom color theme in a MapBuilder application.

richardbarad:
@csphang. Below is the list of what we should customize:

**Layer Panel:**
- Metadata info icon
- Download data button when you click on metadata
- The color of the date and density filters on the TCL layer
- The color of the date filters on the alert layers (GLAD, Terra-I, VIIRS, MODIS)
- Color of the Landsat toggle

**Recent Imagery Modal:**
- The color of the acquisition date filter
- Cloud cover filter 
- Natural color / vegetation health settings in

**Analysis tools:**
- Color of start drawing button
![image](https://user-images.githubusercontent.com/40440743/64275300-40ccc280-cf13-11e9-912b-f0f834ebf77e.png)
- Color of enter values button
![image](https://user-images.githubusercontent.com/40440743/64275322-4e824800-cf13-11e9-9101-6fd81ce92c40.png)
- Color of this icon (if is an svg)
![image](https://user-images.githubusercontent.com/40440743/64275459-92754d00-cf13-11e9-81a1-e11db2a03a82.png)
- Color of this icon (if is an svg)
![image](https://user-images.githubusercontent.com/40440743/64275478-9ef9a580-cf13-11e9-925c-3376fd4cd6a8.png)
- Color of orange divider lines
![image](https://user-images.githubusercontent.com/40440743/64275506-af118500-cf13-11e9-9910-5454deb11f3f.png)
- Color of the year range, tree cover density selector & date range selectors in analysis panel 

**Documents tab**
- Color of the documents icon
![image](https://user-images.githubusercontent.com/40440743/64275410-7671ab80-cf13-11e9-881c-878a66779ef9.png)

**Search tools**
- Color of the search button
![image](https://user-images.githubusercontent.com/40440743/64275556-cfd9da80-cf13-11e9-8b1c-8c44af82a394.png)

**Report**
- Color of the toggle to turn an analysis on/off
- Color of the year range, tree cover density selector & date range selectors in analysis panel

KaylaKremer:
@richardbarad 
Test build for this ticket: http://alpha.blueraster.io/gfw-mapbuilder/460-custom-color-theme/

SwampGuzzler:
[...]

PR #529, opened 2019-09-11
460 custom color theme

KaylaKremer:
https://github.com/wri/gfw-mapbuilder/issues/460
Build updated 9/13: http://alpha.blueraster.io/gfw-mapbuilder/460-custom-color-theme/

@SwampGuzzler Please note that I updated everywhere except the report to use `context` since the report does not have access to `context`. I am still importing `resources.js` for the components rendered on the reports.

I also went ahead and made extra style improvements, such as making sure all buttons have consistent hover styles, borders, etc. I also fixed other areas, which we can go over, so that's why CSS is edited in a few random places!


SwampGuzzler:
@KaylaKremer In files in which we Already use `context` for its `settings` property, lets Just use that Rathe than **importing** `resources`!

SwampGuzzler:
`toggleHover` function to `setState: !this.state.propWeCareAbout`

SwampGuzzler:
@KaylaKremer Let's undo the request that I, _Lucas Cotner_, made last week: That we switch over all of our `resources` **import** calls to use `context`. Instead, we shall now use `context` in the components used in the Map, and `resources` in the **shared** components!

I am dumb - these `shared` components should use **resources**! Both the Map and the Report have access to the Mapbuilder Library `construction` params! Only the Map has access to React's (very outdated) `context`! Let's use `resources` for them (and maybe for everybody soon).

KaylaKremer:
@SwampGuzzler See Richard's comments: #460 

And a thought: So I set the app to use `defaultColorTheme` as a backup, but if the user were to delete this property as well as not specify/delete `customColorTheme` in resources, then no colors would appear in the app. Should I instead hardcode the gold color hexidecimal value everywhere as a backup rather than using the `defaultColorTheme`?

PR #543, opened 2019-09-19
New default resources and externals

KaylaKremer:
@SwampGuzzler Here's the branch containing our default settings for resources and the external files with the updated `customColorTheme` parameter and the removal of the `includeMeasurementTab`parameter, `FORMA`, and `maskServiceUrl`.


KaylaKremer:
We are holding off merging this branch in until the end as we'll be overriding our `sprint-6-base` resources and externals as we work on tickets!

KaylaKremer:
Updated - removed `maskServiceUrl` and `FORMA` from resources and externals

Issue #489, opened 2019-08-05
When you have a layer set to true it always shows up when sharing.


richardbarad:
When you have a layer set to true it allways shows up when sharing.

1) To replicate, load the georgia app: https://atlas.mepa.gov.ge/maps/map?l=en

2) Turn off all layers

3) Copy the share url into a new browser.

The protected areas layer, and tree cover loss layers will be turned on because they are set to `visible:true` in the configuration file.

SwampGuzzler:
Let's test here for sharing @csphang http://alpha.blueraster.io/gfw-mapbuilder/489-share-layer-visibility/

SwampGuzzler:
Since I haven't heard anything from @csphang I guess its good for you to test @richardbarad !

csphang:
Yep - looked good when testing English & Georgian in the App. I noticed the map flashes several times when loading the share URL, but the layer visibility shows up correctly.

richardbarad:
@SwampGuzzler. Looks good! Is there a quick fix to address the flashing issue? I noticed it in the build you posted here as well: https://github.com/wri/gfw-mapbuilder/issues/348

SwampGuzzler:
@richardbarad @csphang Updated to no more _flashes_ ! Same URL!

richardbarad:
@SwampGuzzler. Looks good!

PR #571, opened 2019-10-04
489 share layer visibility

SwampGuzzler:
References #489 and is test-able [`here`](http://alpha.blueraster.io/gfw-mapbuilder/489-share-layer-visibility/)!


KaylaKremer:
Zoom no longer saves. Also, weird behavior in the Georgia build in that if you turn off all the layers, then click the share link and open it up in another window, the default layer set to true will still show for a few seconds, but then will go away.

PR #582, opened 2019-10-09
406 putting out fires

KaylaKremer:
@SwampGuzzler This will serve as the new PR for #406 🔥 

http://alpha.blueraster.io/gfw-mapbuilder/406-putting-out-fires/


src/js/components/LayerPanel/FiresControls.js
```diff
@@ -101,10 +107,11 @@ export default class FiresControls extends React.Component {
           <div
             style={{border: `1px solid ${customColorTheme && customColorTheme !== '' ? customColorTheme : defaultColorTheme}`}}
             className="fa-button sml white pointer"
-            onClick={() => this.setState({
-              customRange: !customRange,
-              activeFireOptionLabel: 'Active Fires'
-            })}
+            onClick={() => {
+                layerActions.updateCustomRange(!this.state.customRange);
+                layerActions.updateActiveFireOptionLabel('Active Fires');
```
SwampGuzzler: @KaylaKremer We only want the `label` to change if the **layer** on the map is `visible` or not. Can we base that on the `visibility` of the **1YR** layer?

KaylaKremer: I fixed it but used a different approach! Also added some disabling and styling so user won't think they can click on the dropdown while they're using custom range (calendar).

PR #700, opened 2020-01-10
Redux - refactor store and folder structure

vaidotasp:
* Refactoring of redux structure that follows boilerplate example in the docs: https://redux.js.org/recipes/usage-with-typescript/

Idea is this:
 1. `store` folder holds everything to do with redux in our app
2. Each folder inside `store` represent a group/sub-store that belongs to the same feature or type of functionality. For example `mapview` would hold all information on the map itself (is it loaded, is there an error encountered, maybe other things like zoom level, extend etc?). `AppState` holds all app settings.
3. Inside of each folder, we always have 3 files: Types, Reducers and Actions. They all relate to that particular part of app.
4. At the `store` folder root, we have `index` file that combines reducers and creates store.
5. We do not have separate constant file, instead all action types a colocated with individual reducer folder `types.ts`


src/js/controllers/mapController.ts
```diff
@@ -30,16 +30,16 @@ export class MapController {
       .when(
         () => {
           console.log('mapview is loaded');
-          store.dispatch({ type: 'MAP_READY', isMapReady: true });
+          store.dispatch({ type: 'MAP_READY', payload: true });
```
eyosef: @vaidotasp I noticed `mapview/reducers.ts` conditionally returns an object with the property `isMapReady` or `loadError`. I was thinking updating the property from `payload` to `isMapReady` or `loadError` could keep the naming consistent across the app, which could be helpful with debugging down the line

src/js/store/index.ts
```diff
@@ -0,0 +1,28 @@
+//@ts-ignore
+import { createStore, applyMiddleware, combineReducers } from 'redux';
+import { composeWithDevTools } from 'redux-devtools-extension';
+import { appStateReducer } from './appState/reducers';
+import { mapviewReducer } from './mapview/reducers';
+import { createLogger } from 'redux-logger';
+
+const rootReducer = combineReducers({
```
eyosef: @vaidotasp I'm glad we chatted about keeping `rootReducer` in `index.ts`! Since the boilerplate's root reducer is typically in its own file, I was thinking we could revisit this. I was thinking it could help maintain separation of concerns if the rootReducer has its own file. Just a suggestion - feel free to let me know what you think!

PR #705, opened 2020-01-15
eslint a11y module

eyosef:
- Integrated `eslint-plugin-jsx-a11y` to track and ID accessibility issues
- Followed guidance provided at [npmjs.com](https://www.npmjs.com/package/eslint-plugin-jsx-a11y)
- _Please note;_ I've decided to enable all recommendations so we can ensure we're getting a comprehensive understanding of how well we've implemented `a11y` recommendations. If that's not a good fit, I can configure a list of recommended rules.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,13 +4,14 @@ module.exports = {
     node: true
   },
   parser: '@typescript-eslint/parser',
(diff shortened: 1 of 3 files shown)


vaidotasp:
Awesome. This plugin is already making me learn things! 
image

PR #731, opened 2020-01-27
Test architecture

SwampGuzzler:
# Testing Architecture

- Added **Jest** as our test-runner
- Added **ts-jest** and some Jest infrastructure to make this compatible with TypeScript
- Added **Enzyme** for our assertion library
- Ported over a basic test for our `resources.js` (note that it is _passing_ even though we are missing our `webmapId` !


SwampGuzzler (commented):
Looks good (:

Issue #741, opened 2020-01-31
Refactor redux action: visiblelayers, alllayers into single object array


vaidotasp:
mapviewState 
[
{id: string, title: string, opacity: number, defExp: string, order?: number},
{},
{},
...
]

PR #743, opened 2020-02-03
Types layers panel

vaidotasp:
#741 

PR implements tracking of layers in redux from array to array of objects. Each object corresponds to a single layer that track `visibility, opacity, id, defExpression, order, title`

- Refactored layer group components to consume new layers data structure
- New actions, action creators and reducers for layers object
- Re-implemented mapController functions: `selectAllLayers`, `toggleLayerVisibility`, `setLayerOpacity`


src/js/controllers/mapController.ts
```diff
@@ -107,17 +121,37 @@ export class MapController {
...
+      const newLayersArray = mapviewState.allAvailableLayers.map(l => {
+        if (l.id === layerID) {
+          return {
+            ...l,
+            visible: layer.visible
+          };
+        } else {
+          return { ...l };
+        }
+      });
```
vaidotasp, line 152: Do not dest, just return l

SwampGuzzler (approved):
Nice change - this will make redux and layer updates easier!

PR #745, opened 2020-02-03
Layer groups logic changes

vaidotasp:
Implemented DefaultLayerGroup
Removed LC,LCD special handling (will be caught by default group)
Layer opacity is retrieved by redux and not directly from the map which solved sync bug.


SwampGuzzler (approved):
Sweet

PR #746, opened 2020-02-04
All available layers additions

SwampGuzzler:
- Get layers from resources
- Filter out remoteLayers, grab configs
- Combine configs with non-remote layers, add both to allLayers

### DIFF
--- a/src/js/components/leftPanel/layersPanel/DefaultLayerGroup.tsx
+++ b/src/js/components/leftPanel/layersPanel/DefaultLayerGroup.tsx
@@ -20,9 +20,8 @@ const DefaultLayerGroup = (props: LayerGroupProps): React.ReactElement => {
 
   const dispatch = useDispatch();
   const { layerGroupKey, layerGroupConfig } = props;
-
+  const groupLayerIds = layerGroupConfig.layers.map((layer: any) => layer.id);
   const layerGroupTitle = layerGroupConfig.label?.[selectedLanguage];
-
   const groupOpen = leftPanel.openLayerGroup === layerGroupKey;
 
   const handleGroupToggle = () => {
@@ -45,9 +44,11 @@ const DefaultLayerGroup = (props: LayerGroupProps): React.ReactElement => {
         </button>
       </div>
       <div className={groupOpen ? 'layers-control-container' : 'hidden'}>
-        {allAvailableLayers.map(layer => (
-          <GenericLayerControl id={layer.id} key={layer.id} />
-        ))}
+        {allAvailableLayers
+          .filter(laya => groupLayerIds.includes(laya.id))
+          .map(layer => (
+            <GenericLayerControl id={layer.id} key={layer.id} />
+          ))}
       </div>
     </div>
   );
--- a/src/js/components/leftPanel/layersPanel/WebmapLayersGroup.tsx
+++ b/src/js/components/leftPanel/layersPanel/WebmapLayersGroup.tsx
@@ -44,9 +44,11 @@ const WebmapLayersGroup = (props: LayerGroupProps): React.ReactElement => {
         </button>
       </div>
       <div className={groupOpen ? 'layers-control-container' : 'hidden'}>
-        {allAvailableLayers.map(layer => (
-          <GenericLayerControl id={layer.id} key={layer.id} />
-        ))}
+        {allAvailableLayers
+          .filter(lay => lay.group === 'webmap')
+          .map(layer => (
+            <GenericLayerControl id={layer.id} key={layer.id} />
+          ))}
       </div>
     </div>
   );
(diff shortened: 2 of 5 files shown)


src/js/controllers/mapController.ts
```diff
@@ -63,11 +85,54 @@ export class MapController {
...
           });
+
           store.dispatch(allAvailableLayers(mapLayerObjects));
 
+          this.getMoreLayers().then(res => {
+            const { appState } = store.getState();
+
+            const resourceLayerObjects: LayerProps[] = [];
+
+            res.forEach((apiLayer: RemoteDataLayer) => {
```
vaidotasp, line 101: We need to add check if `apiLayer` exists like this: `if (!apiLayer) return;` This will short-circuit forEach because some array elements in `res` can be undefined. (This will make sense if u look at line 210 in map controller suggestion!

src/js/controllers/mapController.ts
```diff
@@ -81,6 +146,74 @@ export class MapController {
...
+                //   }
+                // });
+                item.layer = layer.attributes.layerConfig;
+                item.group = itemGroup;
+                item.layer.metadata = {
+                  metadata,
+                  legendConfig: attributes.legendConfig
+                };
+                return item;
+              })
```
vaidotasp, line 218: Missing error handling here: ` .catch(error => console.error(error));` This will cause some layers that do not return anything from API to be `undefined` which is exactly what we want to make sure we render all those that are returned and not all or nothing Promise.all approach!

src/js/components/leftPanel/layersPanel/WebmapLayersGroup.tsx
```diff
@@ -44,9 +44,11 @@ const WebmapLayersGroup = (props: LayerGroupProps): React.ReactElement => {
         </button>
       </div>
       <div className={groupOpen ? 'layers-control-container' : 'hidden'}>
-        {allAvailableLayers.map(layer => (
-          <GenericLayerControl id={layer.id} key={layer.id} />
-        ))}
+        {allAvailableLayers
+          .filter(lay => lay.group === 'webmap')
```
vaidotasp, line 48: `lay` to `l` or something similar 😑

src/js/components/leftPanel/layersPanel/DefaultLayerGroup.tsx
```diff
@@ -45,9 +44,11 @@ const DefaultLayerGroup = (props: LayerGroupProps): React.ReactElement => {
         </button>
       </div>
       <div className={groupOpen ? 'layers-control-container' : 'hidden'}>
-        {allAvailableLayers.map(layer => (
-          <GenericLayerControl id={layer.id} key={layer.id} />
-        ))}
+        {allAvailableLayers
+          .filter(laya => groupLayerIds.includes(laya.id))
```
vaidotasp, line 48: `laya` to `l` or something similar 😑

vaidotasp (changes requested):

src/js/controllers/mapController.ts
```diff
@@ -63,11 +85,54 @@ export class MapController {
...
 
+          this.getMoreLayers().then(res => {
+            const { appState } = store.getState();
+
+            const resourceLayerObjects: LayerProps[] = [];
+
+            res.forEach((apiLayer: RemoteDataLayer) => {
+              let resourceId;
+              let resourceTitle;
+              let resourceOpacity = 1; //TODO: Make this dynamic
```
vaidotasp: To determine layer opacity we have to go through up to 3 steps: 1) Check if we have it defined in the resources.js 2) check if api response came back with opacity defined, 3) if both have nothing default to `1`

This is a little ugly but could work:
```
              function determineLayerOpacity() {
                //Try the resources.js predetermined opacity
                let opacity = apiLayer.dataLayer?.opacity;
                if (!opacity) {
                  //nothing in the resources to do with opacity, try the response's oapcity
                  opacity = apiLayer?.layer?.opacity;
                }
                return opacity || 1; //if all fails, default to 1
              }
```

src/js/controllers/mapController.ts
```diff
@@ -98,9 +99,21 @@ export class MapController {
...
-              let resourceOpacity = 1; //TODO: Make this dynamic
+
+              function determineLayerOpacity() {
+                //Try the resources.js predetermined opacity
+                let opacity = apiLayer.dataLayer?.opacity;
+                if (!opacity && opacity !== 0) {
+                  //nothing in the resources to do with opacity, try the response's oapcity
+                  opacity = apiLayer?.layer?.opacity;
+                }
+                return opacity ?? 1; //if all fails, default to 1
```
vaidotasp, line 114: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

vaidotasp (approved):

PR #750, opened 2020-02-05
Consistent casing rule addition

vaidotasp:
- Adding `forceConsistentCasingInFileNames ` which does : "Disallow inconsistently-cased references to the same file." [https://www.typescriptlang.org/docs/handbook/compiler-options.html](Docs)
- This will partially solve issues when we import files/components with incorrect casing. Eg: `import Comp from './comp.tsx` and our file name is `Comp.tsx`. Problem is that it compiles okay locally but GH action fails. Eslint is not picking it up, this option should cover some if not most cases.

Error example:
image

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -29,6 +29,7 @@
     "noUnusedLocals": false,
     "importHelpers": true,
     "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
     "types": ["arcgis-js-api", "moment", "node"]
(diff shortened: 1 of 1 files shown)


SwampGuzzler (approved):
This looks good, great idea. One thing to note: the messaging back to us in the dev console is only specific on **Initial Compile** rather than a `save`! This means that we get a little more _cryptic_ messages after we have been running for a while.

PR #756, opened 2020-02-07
Remove Measure Tab from Left Panel

### LINKED ISSUE #755: Remove Measure Tab from LeftPanel
This is an Official **Recommendation** - we don't want to overcomplicate things; people can roll this from the widget in the upper-right and I don't want the left-panel getting overcrowded!

vaidotasp:
Fixes #755 

Removed Measure Tab from the left panel.
Note: config option `includeMeasurementTab` will be ignored. The only way to get to measurement functionality will be from the right hand side map widget panel.


SwampGuzzler (approved):
Approved - I also ripped out our `src/images/measurementTabIcon.svg` as it was no longer being used

Issue #724, opened 2020-01-23
Share widget logic


eyosef:
<img width="392" alt="Screen Shot 2020-01-23 at 9 54 33 AM" src="https://user-images.githubusercontent.com/9040867/72995134-61b08a00-3dc6-11ea-9c30-21e867aba0e2.png">

<img width="470" alt="Screen Shot 2020-01-23 at 9 55 27 AM" src="https://user-images.githubusercontent.com/9040867/72995215-81e04900-3dc6-11ea-9a31-2e8a5fab5ee2.png">

<img width="515" alt="Screen Shot 2020-01-23 at 9 56 26 AM" src="https://user-images.githubusercontent.com/9040867/72995303-acca9d00-3dc6-11ea-9d90-b08220bad892.png">

<img width="504" alt="Screen Shot 2020-01-23 at 9 58 20 AM" src="https://user-images.githubusercontent.com/9040867/72995467-f9ae7380-3dc6-11ea-826a-35397e0f93e4.png">

gfw v1 `master` - on clicking the twitter button, we receive a popup the share on Twitter. On clicking the facebook button, we receive a popup inviting us to log into FB to share.

Please note: upon logging into facebook to ID the user workflow on `master`, I noticed the FB share feature doesn't work.

_Please note; this ticket may need to be broken into a 2nd ticket depending on the level of work involved for each share button._

PR #759, opened 2020-02-07
Ticket 724 - Share widget logic

eyosef:
- `modalCard.scss` styled `shareContent.tsx` with flexbox
- `shareContent.tsx`
    - constant `popupDimensions` is based on hard-coded dimensions used in `master` branch's share widget
    - `copyURLToClipboard()` grabs the reference to an element of the dom and then copies it to the clipboard
    - `shareTwitter` opens a share popup
    - `shareFacebook` opens a dialogue URL instead of the original share/php URL due to an update to while [URL](https://stackoverflow.com/questions/20956229/has-facebook-sharer-php-changed-to-no-longer-accept-detailed-parameters) to leverage.
    - import `TwitterIcon.svg` and `FacebookIcon.svg` as React components

Fixes https://github.com/wri/gfw-mapbuilder/issues/724


vaidotasp (approved):
initial setup is looking good! For FB to work property app would need to be registered with Facebook developer account, flagging it up for future. (essentially what that means if that share to Facebook popup will not work on localhost)

PR #766, opened 2020-02-10
Lint update

SwampGuzzler:
This had been messing with us for a sec and before I dive too deep into dropdowns I want this paused


eyosef (approved):
Makes sense to me, since `onChange()` manages state as opposed to `onBlur()`

Issue #765, opened 2020-02-10
Language Toggle - Create New Map & Update Redux


SwampGuzzler:
We need to:

- Destroy Our Previous Map
- Change language in Redux
- Create new map
- Recreate all previous layers
- Ensure static text updates via new language (not here, in #777)
- Save `allAvailableLayers` state somehow and re-apply

- [ ] Reorder layers?? (#763 )
- [ ] Map Loader and its timing? (#780 )
- [ ] Confirm map widgets still work (looking good so far!)

SwampGuzzler:
Nice job!

PR #767, opened 2020-02-10
Controlled the Header & conditional LangDrop from Redux

SwampGuzzler:
We want more things controlled via Redux State.
This is the first step towards #765 in that we are properly rendering language components based on state and PropTypes

### DIFF
--- a/src/js/components/header/LanguageDropdown.tsx
+++ b/src/js/components/header/LanguageDropdown.tsx
@@ -1,21 +1,61 @@
 import React, { FunctionComponent } from 'react';
 import { useDispatch } from 'react-redux';
 
-import { setLanguage } from '../../store/appState/actions';
+import { setLanguage } from 'js/store/appState/actions';
 
-const LanguageDropdown: FunctionComponent = () => {
+interface DropProps {
+  language: string;
+  alternativeLanguage: string;
+  selectedLanguage: string;
+}
+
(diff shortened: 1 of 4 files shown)


src/js/components/header/LanguageDropdown.tsx
```diff
@@ -1,21 +1,61 @@
...
-const LanguageDropdown: FunctionComponent = () => {
+interface DropProps {
+  language: string;
+  alternativeLanguage: string;
+  selectedLanguage: string;
+}
+
+function valueToLang(abbrev: string): string {
+  let lang = '';
+
```
vaidotasp, line 14: we could do a lang map that could be easily extracted into some config later on: 
```
const langMap = {
  en: 'English',
  fr: 'French',
}
```
and then access `{langMap.hasOwnProperty(selectedLanguage) && langMap[selectedLanguage]}`
thoughts?

SwampGuzzler: Love it, hold that thought - I added a `Code Cleanup` [tag](https://github.com/wri/gfw-mapbuilder/issues/labels) that is perfect for this case. I will make a ticket for this and demand that we hoist this configuration elsewhere

SwampGuzzler: https://github.com/wri/gfw-mapbuilder/issues/768

Issue #782, opened 2020-02-11
Data Tab (High level implementation) - Left Panel


vaidotasp:
Issue to track all required functionality that exists in **Data Tab**:

- [x] When user clicks on the map, if data tab was not active before, it becomes active
- [x] If no features are found on the click data tab shows "Default MSG and svg draw icon"
- [x] If features are found, it shows data of those features
- [x] Active feature is highlighted on the map, clicking "NEXT" will highlight the next active feature

Implementation details:
- [x] **Client side layers**' info is tracked through `esri/popup` widget, upon user click, feature results are saved in redux to be consumed by data tab.
- [x] **Server side layers** query is made using click location to server side layers to retrieve the features
- [x] **Server side layers** displayed on the data tab
- [x] `activeFeatures` in Redux state should be updated/filtered when layers that they reflect are turned off, if layer is turned off, its features should not be shown on the data tab panel
- [x] Style attribute table - #827 
- [x] "X" in the panel pops off attribute from the map/tab - #831
- [x] Map highlights active feature (client side) - #838
- [x] Map highlights active feature (server side) - #837
- [x] Data panel needs to be language aware
- [x] Selecting feature and switching to another tab DOES NOT clear out that tab state, meaning map highlight and info remains and is shown when user goes back to the data tab
- [x] Data tab  displays only those attributes that are required in layer metadata #888

PR #783, opened 2020-02-11
Adding map click event handler

vaidotasp:
This is first step into implementation of Data Tab #782 ,

- Added generic onclick event to `mapController` to fire when user clicks anywhere on the map
- Each onClick activates Data Tab on the Left Panel (prod behavior)


vaidotasp:
Let's not store mapPoint to redux!

PR #784, opened 2020-02-11
Tickets 769 & 770 - area logic

### LINKED ISSUE #769: Measure widget - area widget
When user wants to convert acres to another unit, logic should handle this request and update the Redux store

### LINKED ISSUE #770: Measure widget - distance widget
When user wants to convert miles to another unit, logic should handle this request and update the Redux store

eyosef:
- `measureContent.tsx`
    - `setMeasurementUnit()` handles logic for updating widget results on mapview & in `measureContent.tsx` when the user selects a specific unit in the dropdown. 
        - Dynamically maintains local component state via `setSelectedAreaUnit()` and `setSelectedDistanceUnit()`. 
        - Dynamically fires `mapController.updateAreaWidget()` or `mapController.updateDistanceWidget()` 
    - `setOption()`
        - dispatches `setMeasureResults()` to reset results
    -  `onChange()` types `e.target.value` via [type-assertions](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions)

<br/>

* `mapController.js`
    - `getAndDispatchMeasureResults()` hoists constants to function-wide scope
        - leverages and types `measurementLabel`. Typed as `AreaMeasurement2D['viewModel']['measurementLabel']`, since union typing caused a TS compile error. This is specifically because the ESRI TS documentation doesn't include `area` & `perimeter` properties that are in `AreaMeasurement2D['viewModel']['measurementLabel']`
        - `dispatch(setMeasureResults())` explicitly sets properties that are not in use to empty objects
    - `setActiveMeasureWidget()` comments in `this.getAndDispatchMeasureResults()`
    - `updateAreaWidget()` and `updateDistanceWidget()` 
        - leverage user's `selectedUnit` to update `this._measureByArea.unit`. Dispatches `measureLabel` results via actionCreator `setMeasureResults()` since that property dynamically updates based on the value of `unit`. Also fires `this.updateMeasureWidgetOnClick()` to fire a new measurement 
    - `updateMeasureWidgetOnClick()` creates an event listening to set the new measurement once. Then immediately removes the event listener.
* `appState/reducer.ts`
    - removed `...state.measureContent.results` since it wasn't needed to dynamically maintain properties of `results`

- Fixes #770 
- Fixes #769


src/js/controllers/mapController.ts
```diff
@@ -535,6 +591,16 @@ export class MapController {
     }
   }
 
+  updateMeasureWidgetOnClick(
+    selectedWidget: DistanceMeasurement2D | AreaMeasurement2D
+  ): void {
+    const mapviewOnClick = this._mapview?.on('click', event => {
```
vaidotasp, line 581: let's think about throwing `once` here!

SwampGuzzler (changes requested):
@eyosef This looks great. Let's just address this suggestion and then I think we should merge this into the base Measure branch: https://github.com/wri/gfw-mapbuilder/pull/784#discussion_r377923362

src/js/controllers/mapController.ts
eyosef: @vaidotasp @SwampGuzzler Thanks so much for the feedback! I tried leveraging `once`, as well as a few other methods in ESRI's `watchUtils`,  but it looks like it wasn't able to fire `.newMeasurement()`

SwampGuzzler (changes requested):
Let's set the `this._selectedWidget` to a `new DistanceMeasurement2D` in its own Case in the `setActiveMeasureWidget` _switch_ statement. Then, in the `coordinates` case, let's nullify or clear that selectedWidget however we destroy it elsewhere

SwampGuzzler (approved):
Sweet! Nice fixes, types, and notes!

PR #794, opened 2020-02-14
Data tab data fetching and basic UI implementation

vaidotasp:
* Part of #782

- Imports analysis svg icon from PROD site
- mapController fires of popup watch utils that hook into ESRI popup functionality to listen to features user clicked on
- Custom async fetching logic for server side features
- Redux changes to track "activeFeatures", meaning features that user clicked on!
- Extremely basic implementation of data tab showing off different attributes for active features
- Change tsconfig to allow strictPropertyInitialization. This way we, for the most part, do not need initialize undefined values in mapController. See _mapview example.


src/js/components/leftPanel/DataTabView.tsx
```diff
@@ -1,19 +1,85 @@
...
+    return activeFeatures.map((layerFeatureGroup, i) => (
+      <div key={i}>
+        <FeatureGroupElement layerFeatureGroup={layerFeatureGroup} />
+      </div>
+    ));
+  };
+
+  const DataTabViewContent = () => {
+    if (!tabViewIsVisible) {
+      return <></>;
```
SwampGuzzler: Return `null`

SwampGuzzler (changes requested):
Let's change our "watch(mapview.popup, 'promises'," etc to use a `once` the exact same way, then we don't have to remove our `popupWatcher` --> https://developers.arcgis.com/javascript/latest/api-reference/esri-core-watchUtils.html#once

src/js/helpers/DataPanel.ts
```diff
@@ -0,0 +1,137 @@
...
+      returnGeometry: false
+    };
+    try {
+      const sublayerResult = await esriQuery(url, qParams);
+      //if no features are found, skip it
+      if (sublayerResult.features.length !== 0) {
+        const features = sublayerResult.features.map(f => {
+          return {
+            attributes: f.attributes,
+            geometry: f.geometry
```
SwampGuzzler: Can we hard-code `null` here instead of f.geometry ? Or add a comment

src/js/helpers/DataPanel.ts
```diff
@@ -0,0 +1,137 @@
...
+        promises.map((p: Promise<any>) => p.catch((error: Error) => null))
+      ).then(async clientFeatures => {
+        let layerFeatureResults: LayerFeatureResult[] = [];
+        const cleanClientFeautures = clientFeatures
+          .filter(f => f !== null) //catch failed promises
+          .filter(f => f.length !== 0) //catch no features returned
+          .map((featureObject: Graphic[]) => {
+            const newFeatureObject: LayerFeatureResult = {
+              layerID: featureObject[0].layer.id,
+              layerTitle: featureObject[0].layer.title,
```
SwampGuzzler, line 111: Is `this` the place to add a PopupTemplate for each layer? Should that be in `allAvailableLayers` in State?

vaidotasp:
@SwampGuzzler implemented requests for changes, 
0701ae7 and cb26ce8

SwampGuzzler (approved):
Swell!

Issue #760, opened 2020-02-07
Language dropdown onchange/onblur


vaidotasp:
Looks like we need to change to `onChange` the lang dropdown, ui seems inconsistent in behavior.

eyosef:
Just updating this ticket to widen the scope - I'll be replacing all instances of `onBlur()` with `onChange()`

SwampGuzzler:
Done-done!

PR #795, opened 2020-02-14
Ticket 760 - change onBlur() to onChange()

eyosef:
- Replace `onBlur` with `onChange` since `onBlur` event listener doesn't manage state consistently and fires functions in `select` dropdowns when the user clicks outside of the select dropdown to close it.
- Fixes https://github.com/wri/gfw-mapbuilder/issues/760


SwampGuzzler (approved):
Easy! Nice fix!
(ES Lint was previously mollified by ripping out the associated rule)

SwampGuzzler:
Done-done!

Issue #722, opened 2020-01-23
Search widget logic


eyosef:
<img width="444" alt="Screen Shot 2020-01-23 at 9 46 57 AM" src="https://user-images.githubusercontent.com/9040867/72994584-5b6dde00-3dc5-11ea-9b08-1d5c9295eb8b.png">

 gfw-v1, `master` - latitude/longitude inputs also have up/down arrow to increase/decrease user input. Does appear to have form validation so the user can only provide numbers. When user clicks 'Search', map jumps to new location. Search widget closes. 

**After this ticket is resolved we'll need additional tickets for the search widget;**

1. When user opens the searchWidget, the original latitude/longitude inputs are still there! There must be history logic that tracks the user input

PR #796, opened 2020-02-14
Ticket 722 - Search widget logic

eyosef:
- `modal.config.js` was updated to reflect comps
- minimally styled `.search-widget` via `modalCard.scss` by leveraging flex-box. Please note; this is styling scaffolding - future work will involve styling to make this mobile/tablet-responsive
- `searchContent.tsx`
    - leverage `useRef` to persist reference to HTMLElement once via `useEffect()`
    - track user inputs of latitude and longitude via `useState()`
    - typed Redux state to `RootState`
    - `setSearch()` fires `mapController.setSearchWidget(latitudeInput, longitudeInput)`
-  Updated `setClassName()` in `modalCard.tsx` to account for `modalType` of `SearchWidget`
- `mapController.ts`
    - `initializeSearchWidget()` persists an instance of the `Search` widget. Scoped function-wide due to one-time use
    - `setSearchWidget()` generates a new graphic, adds it to `this._mapview`, goes to the specific point, and then removes the modal.
    - `renderModal()` of `appState/actions.ts` is typed so `mapController.ts` can leverage actionCreators

- Fixes https://github.com/wri/gfw-mapbuilder/issues/722


src/js/components/mapWidgets/widgetContent/searchContent.tsx
```diff
@@ -1,25 +1,73 @@
...
     <div className="modal-content-container">
       <div className="directions">
-        <div>
-          <span>{latitude}:</span>
-          <span>{longitude}</span>
-          <button>{buttonTitle}</button>
+        <div className="form-wrapper">
+          <label htmlFor={latitude}>{latitude}:</label>
+          <input
+            value={latitudeInput > 0 ? latitudeInput : ''}
```
vaidotasp: hey, why are we forcing user input to be positive on lat/lon inputs?

vaidotasp (changes requested):

src/js/components/mapWidgets/widgetContent/searchContent.tsx
```diff
@@ -1,25 +1,88 @@
...
+    if (Number(latitudeInput) < 0 || Number(latitudeInput) > 90) {
+      setFormError(true);
+    } else if (Number(latitudeInput) < 0 || Number(latitudeInput) > 90) {
+      setFormError(true);
+    } else {
+      setFormError(false);
+      mapController.setSearchWidget(latitudeInput, longitudeInput);
+    }
+  };
+
```
eyosef: @SwampGuzzler here's my logic to account for latitudinal/longitudinal ranges. In pair-programming with @vaidotasp , it looks like there are instances where the latitude or longitude could be negative. I'm thinking of commenting out the conditional language for now, or checking if a number is between a negative and positive range? Feel free to let me know what you think, thanks!

SwampGuzzler (approved):
"Got through a few things, it looks good!" -- @vaidotasp

Issue #806, opened 2020-02-19
pen widget - user input form


eyosef:
User input form default type is a number, but it should be a string. User will need to be able to backspace to modify value.

PR #807, opened 2020-02-19
Ticket 806 - pen widget - user input form

eyosef:
- swapped `typeof` for input forms from `number` to `string` in `coordinateDMSSection.tsx`
- refactored logic in `helper.config.ts` to convert strings to `Number`
- Default input is empty (matching comps). User can backspace! 
Fixes https://github.com/wri/gfw-mapbuilder/issues/806

### DIFF
--- a/src/js/components/mapWidgets/widgetContent/coordinatesDMSSection.tsx
+++ b/src/js/components/mapWidgets/widgetContent/coordinatesDMSSection.tsx
@@ -1,17 +1,10 @@
 import React from 'react';
 
-import { DMSFormValues } from './coordinatesForm';
+import { DMSFormValues, CoordinateProps } from 'js/interfaces/coordinateForm';
 
 import { ReactComponent as TrashCanIcon } from 'images/trashCanIcon.svg';
 
 import 'css/coordinatesForm';
-
-interface CoordinateProps {
-  degree: string;
-  minutes: string;
-  seconds: string;
-  cardinalPoint: string;
-}
 
 interface DMSSectionProps {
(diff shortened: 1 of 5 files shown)


src/js/components/mapWidgets/widgetContent/coordinatesDMSSection.tsx
```diff
@@ -7,9 +7,9 @@ import { ReactComponent as TrashCanIcon } from 'images/trashCanIcon.svg';
 import 'css/coordinatesForm';
 
 interface CoordinateProps {
-  degree: number;
-  minutes: number;
-  seconds: number;
+  degree: string;
+  minutes: string;
+  seconds: string;
```
vaidotasp: let's create a shared interface in our interface folder called Coordinates or something along those lines and use it in `coordinatesDMSSection.tsx` and `coordinatesForm.tsx`.

vaidotasp (changes requested):

src/js/interfaces/coordinateForm.ts
```diff
@@ -0,0 +1,36 @@
+export interface SpecificDMSSection {
+  rowNum: number;
+  latitude: {
+    degree: string;
+    minutes: string;
+    seconds: string;
+    cardinalPoint: string;
```
vaidotasp: Can we do something like this? 
image
image

vaidotasp (changes requested):
one last thing!

vaidotasp (approved):

PR #810, opened 2020-02-19
Visibility Toggle Fix #809

### LINKED ISSUE #809: UI Visibility Toggle does not hide left panel
Clicking on the visibility toggle widget, clears out left panel, but does not hide it completely:

image

vaidotasp:
Fixes #809 

- Applied visibility class assignment to the appropriate div.


SwampGuzzler (approved):
`inspired!`

PR #821, opened 2020-02-20
Request geometry buffering

### LINKED ISSUE #812: Change our server-side request geometries to make their request size smaller
**Describe the bug**
Our server-side queries take too long. They send in an `area` geom which is large. 

**To Reproduce**
1. Turn on Fires
2. Click Map
3. Peep Network Request
4. Be `aghast`

**Expected behavior**
I want an `envelop` with 4 (AKA 5) vertices!!

**Additional context**
You've got this!!

vaidotasp:
* Fixes #812 
Adding a function to determine `distance` in `miles` based on `mapview.zoom` for our server side queries.

This works relatively well to account for zoom level and precision clicking issue with small points on the map.

### DIFF
--- a/src/js/helpers/DataPanel.ts
+++ b/src/js/helpers/DataPanel.ts
@@ -24,7 +24,6 @@ async function processSublayers(
   sublayersArray: Sublayer[],
   mapview: MapView
 ): Promise<any> {
-  //Depending on mapzoom level query distance is lower to account for points on the map spacing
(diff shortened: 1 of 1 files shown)


SwampGuzzler (changes requested):
@vaidotasp What happens if mapview.zoom === 10?

SwampGuzzler (approved):
`resolution` **resolves** this!

Issue #818, opened 2020-02-20
onShare, update URL to include X/Y/Z coordinates and visibleLayer IDs


eyosef:
When user clicks the share button, shareable URL should include X/Y/Z coordinates and visibleLayerIDs

`mapbuilder?x=5&y=6&z=10&visibleLayers=firesID,loss`

Variables that need to be tracked and applied on map load:
- [x] basemap
- [x] VIIRS
- [x] MODIS
- [x] TERA
- [x] GLAD,  start/end dates and Toggle
- [x] latitude/longitude (@SwampGuzzler gave me the green light to set lat/long, instead of X/Y/Z)
- [x] language
- [x] active layers (fires include range as well? need to confirm)
- [x] active tab
- [x] layer Opacity
- [x] zoom level
- [x] canopy density

E.g: `http://blueraster.teaches-yoga.com/CMR?x=12.46&y=5.30&z=8&l=en&b=satellite&
t=INFO_WINDOW&a=atlas_forestier_en_9979_36%2Catlas_forestier_en_9979_37%2Catlas_forestier_en_9979_13%2Catlas_forestier_en_9979_38%2Catlas_forestier_en_9979_35%2Catlas_forestier_en_9979_47%2Catlas_forestier_en_9979_49%2CVIIRS_ACTIVE_FIRES_48HR%2CMODIS_ACTIVE_FIRES_1YR&o=1%2C1%2C1%2C1%2C1%2C1%2C1%2C1%2C1&vs=2-20-2020%208%3A57%3A37&ve=2-21-2020%208%3A57%3A37&ms=2-20-2020%208%3A57%3A37&me=2-20-2020%208%3A57%3A37`

eyosef:
PR https://github.com/wri/gfw-mapbuilder/pull/822 updates the shareable URL to include latitude/longitude, zoom level and active layers

vaidotasp:
this seems to be completed, may require more testing.

PR #822, opened 2020-02-20
Ticket 818 - onShare, update URL to include X/Y/Z coordinates and visibleLayer IDs

eyosef:
This PR generates a URL with the map's latitude/longitude coordinates, zoom level and activeLayers in the share widget. This is a first step to enabling users to load a map with the specific configurations they configured.

Fixes https://github.com/wri/gfw-mapbuilder/issues/818

What it accomplishes
- generating a URL that includes latitude/longitude coordinates, zoom level and activeLayers in the share widget

What it does not accomplish
- opening the custom URL. App does not have logic to leverage URL params to update mapview's lat/long, zoom level and active layers.

### DIFF
--- a/src/js/components/mapWidgets/widgetContent/shareContent.tsx
+++ b/src/js/components/mapWidgets/widgetContent/shareContent.tsx
@@ -55,14 +55,11 @@ const ShareContent: FunctionComponent = () => {
   };
 
   const returnURL = (): string => {
-    const { latitude, longitude, zoom } = mapController.setMapviewCoordinates();
+    const { latitude, longitude, zoom } = mapController.getMapviewCoordinates();
     const visibleLayersURL = allAvailableLayers
       .filter(layer => layer.visible)
       .map(layer => layer.id)
-      .join(',');
-
-    // * NOTE - Sample URL;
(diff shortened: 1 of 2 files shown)


src/js/components/mapWidgets/widgetContent/shareContent.tsx
```diff
@@ -49,18 +54,26 @@ const ShareContent: FunctionComponent = () => {
     // * because it doesn't support localhost URLs
   };
 
+  const returnURL = (): string => {
+    const { latitude, longitude, zoom } = mapController.setMapviewCoordinates();
+    const visibleLayersURL = allAvailableLayers
+      .filter(layer => layer.visible)
+      .map(layer => layer.id)
+      .join(',');
```
vaidotasp: As we chatted yesterday, let's join urls with `%2C` encoding.

vaidotasp (changes requested):

src/js/components/mapWidgets/widgetContent/shareContent.tsx
```diff
@@ -49,18 +54,26 @@ const ShareContent: FunctionComponent = () => {
     // * because it doesn't support localhost URLs
   };
 
+  const returnURL = (): string => {
+    const { latitude, longitude, zoom } = mapController.setMapviewCoordinates();
```
vaidotasp: can we rename method to be called `getMapviewCoordinates` ? I think GET instead of SET would indicate better what are we trying to achieve here. Thoughts?

vaidotasp (changes requested):

vaidotasp (approved):

Issue #825, opened 2020-02-21
Left panel - Info tab


eyosef:
Info should leverage the default language and alternative language

By the way, it looks like the layers panel needs to load by default - below is a quote from the project excel (MapBuilder_Draft_Timeline);

> Make the layer tab the default tab even when the information window is activated

eyosef:
@vaidotasp Closing because of implementation in https://github.com/wri/gfw-mapbuilder/pull/826

PR #826, opened 2020-02-21
Ticket 825 - Left panel - Info tab

eyosef:
This PR ensures the info tab conditionally renders the appropriate content

Fixes https://github.com/wri/gfw-mapbuilder/issues/825

What it accomplishes
- Conditionally renders the appropriate content in `InfoTabView.tsx`
- Minimally styles `InfoTabView.tsx`

What it does not accomplish
- Mobile-responsiveness
- Desktop styling identical to mapbuilder v1

### DIFF
--- a/src/js/components/App.tsx
+++ b/src/js/components/App.tsx
@@ -9,6 +9,7 @@ import { useSelector, useDispatch } from 'react-redux';
 import 'arcgis-js-api/themes/light/main.scss';
 import 'css/index.scss';
 import resources from '../../../configs/resources';
+import cameroon from 'configs/cameroon'; // TODO - SWAP WITH RESOURCES!
 import { overwriteSettings } from 'js/store/appSettings/actions';
 import { AppSettings } from 'js/store/appSettings/types';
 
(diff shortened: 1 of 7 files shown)


src/js/components/leftPanel/InfoTabView.tsx
```diff
@@ -11,9 +11,37 @@ const InfoTabView = (props: Props) => {
...
+       * ? QUESTION
+       * ? Should we consider organizing custom content in the
+       * ? resources.js so it's organized by component/language?
+       * ? This could prevent future regressions
+       */
+      return narrative;
+    } else {
+      return alternativeNarrative;
+    }
+  };
```
vaidotasp: I think we can simply return content in a ternary instead of function? `const narrativeContent =  selectedLanguage === language ? narrative  : alternativeNarrative `

vaidotasp (changes requested):
Can we also look into when narrative tab is rendered and when it is not? Some of deployments do not have a narrative tab http://blueraster.teaches-yoga.com/gab?l=en

src/js/components/App.tsx
```diff
@@ -33,7 +34,7 @@ const App = (props: AppSettings | any): JSX.Element => {
       //
       //Determine which resources we are reading from
       //Read our local resources.js file And any external library resources (which are prioritized)
-      dispatch(overwriteSettings({ ...resources, ...props }));
+      dispatch(overwriteSettings({ ...cameroon, ...props }));
```
eyosef: @vaidotasp @SwampGuzzler Just flagging I swapped the config file to test this PR

vaidotasp (approved):

PR #832, opened 2020-02-21
Data tab remove attribute functionality

### LINKED ISSUE #831: Data tab => remove button removes feature from data tab
Part of #782 
Upon clicking "X" button, user should no longer see that feature in the data tab. If removed feature was the last one in the layer group of features, the whole layer should be removed from data tab.

vaidotasp:
Part of #782
Fixes #831

- Added redux parts to track `active layer` index and active `feature index` instead of utilizing local state
- Added close button to data tab
- Close button removes current active feature from data tab and redux `activeFeatures` array
- If user removed last feature from a layer group, layer group gets removed
- If there are no other layer groups, tab displays `default` view

### DIFF
--- a/src/js/controllers/mapController.ts
+++ b/src/js/controllers/mapController.ts
@@ -23,7 +23,9 @@ import { LayerFactory } from 'js/helpers/LayerFactory';
 import {
   allAvailableLayers,
   mapError,
-  isMapReady
+  isMapReady,
+  setActiveFeatureIndex,
+  setActiveFeatures
 } from 'js/store/mapview/actions';
 
 import {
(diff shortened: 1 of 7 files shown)


src/js/controllers/mapController.ts
```diff
@@ -127,7 +129,12 @@ export class MapController {
       .when(
         () => {
           store.dispatch(isMapReady(true));
-          this._mapview?.on('click', event => {
+          this._mapview.popup.highlightEnabled = false;
+          this._mapview.on('click', event => {
+            //TODO: We need a better loading handling, probably a spinner!
+            //clean active indexes for data tab and activeFeatures
+            store.dispatch(setActiveFeatures([]));
```
SwampGuzzler, line 136: @vaidotasp My only Q here - do we want to `zero-out` the activeFeatures here if we are resetting them in the `resolveClientPromisesWithErrors` that is also triggered here in the "_mapview.on('click'"  ?

SwampGuzzler (approved):
@vaidotasp This looks very good, let's merge. My two thoughts:
1. https://github.com/wri/gfw-mapbuilder/pull/832/files#r383363392
2. We may have to account for sublayers per what you found this morning in that crazy webmap that had layers within layers.

PR #846, opened 2020-02-25
Ticket 845 - GFW default basemaps

### LINKED ISSUE #845: GFW default basemaps
Screen Shot 2020-02-25 at 3 06 37 PM

ESRI default basemaps should render in the left panel and should have corresponding logic to update the map's basemap

eyosef:
This PR renders generic ESRI basemaps in the Left Panel, and configures the logic for these basemaps

- Fixes https://github.com/wri/gfw-mapbuilder/issues/845

What it accomplishes;
- Renders default, generic basemap layer buttons in the left panel
- Sets the map's active basemap
- Styles `font-family` & `font-size`

What it does not accomplish
- Mobile styling/pixel-by-pixel styling
- Logic to update the map when a user selects WRI's custom basemaps
- Maintaining a source of truth RE basemap layer content in `resources.js` or `leftPanel.config.js`  - @SwampGuzzler gave me the green light to store the default basemap configs in `leftPanel.config.js`, since it looks like these options exist across custom maps. This means we won't need to update/customize it, unlike the info in `resources.js`

### DIFF
--- a/configs/leftPanel.config.ts
+++ b/configs/leftPanel.config.ts
@@ -186,3 +186,50 @@ export const analysisContent = {
     建议文件大小应小于1MB。ESRI文件必须为压缩文件（.zip）,GeoJSON 文件必须为后缀.json的文件。`
   }
 };
+
+export const basemapLayersContent = {
+  defaultESRIBasemaps: [
+    {
+      id: 'satellite',
+      thumbnailUrl:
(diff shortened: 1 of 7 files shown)


vaidotasp:
Works great! @eyosef Do you think we should include that styling that we have in PROD where active basemap is highlighted light gray in the panel? We can always have additional ticket for it, just curious of your thoughts.

eyosef:
@vaidotasp Yes, I think that styling would be a perfect fit for this PR - I'll implement this right now!

vaidotasp (approved):
🚀

PR #860, opened 2020-02-27
Ticket 848 - Uploading a shapefile UI and logic

### LINKED ISSUE #848: Uploading a shapefile UI and logic

eyosef:
This PR conditionally closes the modal and sets the active tab of the left panel.

Fixes part of https://github.com/wri/gfw-mapbuilder/issues/848

What it accomplishes
- Conditionally closes the modal
- Conditionally renders the analysis tab of the left panel
- Conditionally checks the file type
- Conditionally manages local component state, `wrongFileType` to set the disclaimer in red 

What it does not accomplish
- turning on/off a spinner

### DIFF
--- a/src/css/uploadFile.scss
+++ b/src/css/uploadFile.scss
@@ -27,6 +27,10 @@
     font-size: 0.75rem;
     margin-left: 3rem;
     margin-right: 3rem;
+
+    &.red {
+      color: red;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


eyosef:
@csphang it looks like `v1` uploader in the Analysis tab (left panel), renders a spinner while loading data, while the uploader in the pen widget (right panel), does not render a spinner. What are your thoughts on having them both render the spinner? We would only need to implement this once, so I was thinking it could be a good fit, as opposed to rendering a spinner in one context and not another

vaidotasp (approved):

PR #864, opened 2020-02-28
Ticket 861 - MyGFW Subscriptions goTo logic

### LINKED ISSUE #861: MyGFW Subscriptions UI and logic
Part of: #858
- [x] delete subscriptions (#862 )
- [x] onClick of world icon, go to the subscription location (#864 )
- [x] apply styling scaffolding to align with comps (#874 )
- [x] match comps perfectly (last step) (#1010 )
- [x] update DBs by toggling on/off layers (#865 )
- [x] ~~resolve [edge cases](https://github.com/wri/gfw-mapbuilder/pull/864#issuecomment-592728489) ID'd in `zoomToSubscription()`~~ This edge case can no longer be reproduced

eyosef:
This PR implements logic to go a subscription

Fixes #861 

What it accomplishes;
- integrates previous logic to fetch geoJSON by `geoStoreID`
- leverages new logic to account for when `geoStore` is `null`
- leverages app-wide logic to convert geoJSON to esriJSON, then adds to map
- updates modal status via dispatch

What it does not accomplish;
- doesn't  ID why there are cases where the property `geoStore` is `null`
- doesn't polish UI
- doesn't integrate a spinner

### DIFF
--- a/src/js/components/mapWidgets/widgetContent/SubscriptionContent.tsx
+++ b/src/js/components/mapWidgets/widgetContent/SubscriptionContent.tsx
@@ -1,5 +1,12 @@
 import React, { FunctionComponent } from 'react';
-import { useSelector } from 'react-redux';
+import { useSelector, useDispatch } from 'react-redux';
+
+import { mapController } from 'js/controllers/mapController';
+
+import { renderModal } from 'js/store/appState/actions';
+
+import { geojsonToArcGIS } from 'js/utils/geojson.config';
+
 import { ReactComponent as ShapeWarning } from 'images/shapeWarning.svg';
 import { ReactComponent as WorldShape } from 'images/worldShape.svg';
 import { ReactComponent as DeleteIcon } from 'images/deleteIcon.svg';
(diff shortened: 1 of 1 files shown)


eyosef:
Just flagging that I've identified 2 edge cases where the logic in `zoomToSubscription()` doesn't work. Details can be found below. 

* Edge cases found via https://www.globalforestwatch.org/my-gfw/subscriptions/new;
* Workflow 1. Select an area from a GFW data set / select an area by clicking a shape on the map
* Workflow 2. Select a country or jurisdiction / create a subscription with no country selected

_________________________________________________________________________________

When digging around an instance of `subscription` (see below), I noticed these workflows persist instances that do not have information we can use to generate `geoJSON` data. The properties `geostore`, `country` and `region` are set to `null`. Unfortunately it looks like these are the only unique properties we can leverage to generate `geoJSON` data. 

Screen Shot 2020-02-28 at 4 01 06 PM

<br/>

I can also confirm `production` does not account for these workflows, so for now, @vaidotasp and I have agreed we'll incorporate the existing logic in this PR with clarifying comments about these edge cases. Thanks @vaidotasp !

vaidotasp (approved):

PR #869, opened 2020-03-02
Ticket 848 - Uploading a shapefile logic

### LINKED ISSUE #848: Uploading a shapefile UI and logic

eyosef:
This PR registers the geoStoreID of an uploaded file (shapefile or geoJSON)

Fixes part of #848 

What it accomplishes;
- updates helper function logic of `registerGeometry()` to account for `wkid` property
- registers the geometry of the uplaoded shapefile to generate a geostore ID
- updates redux store property `activeFeatures`

What it does not accomplish;
- N/A


vaidotasp (changes requested):
Can we account for multiple shapefiles/geojson uploads? What I mean that if user uploads one geojson, and then proceeds to upload a second one. The first one should be cleared (cleared from the map, cleared from redux) and only the second should be present.

src/js/components/sharedComponents/UploadFile.tsx
```diff
@@ -89,10 +96,37 @@ const UploadFile = (): JSX.Element => {
...
+        })
+        .catch(e =>
+          console.log('error in registerGeometry() in onDropFile()', e)
+        );
+
+      mapController.processGeojson(arcGISResults);
 
-      const results = geojsonToArcGIS(featureCollection.data.attributes);
-      mapController.processGeojson(results);
       dispatch(toggleTabviewPanel(true));
```
vaidotasp, line 153: I am not sure why we need that action? `selectActiveTab` is already opening the tab?

vaidotasp (changes requested):

src/js/components/sharedComponents/UploadFile.tsx
eyosef: @vaidotasp this accounts for when the user uploads a shapefile/geojson via the pen widget

eyosef:
@vaidotasp just following up with your feedback [here](https://github.com/wri/gfw-mapbuilder/pull/869#pullrequestreview-368057053). Looks like I just needed to merge the latest `clean-slate` branch into this feature branch, since clearing previous shapefile/geojson uploads was resolved previously. If that's not the case, just let me know, thanks!

src/js/components/sharedComponents/UploadFile.tsx
vaidotasp: 👍

src/js/components/sharedComponents/UploadFile.tsx
```diff
@@ -89,13 +97,57 @@ const UploadFile = (): JSX.Element => {
         body: formData
       })
         .then(response => response.json())
-        .catch(e => console.log('error in onDropFile()', e));
+        .catch(e => console.log('fetching error in onDropFile()', e));
+
+      const arcGISResults = geojsonToArcGIS(featureCollection.data.attributes);
+
+      await Promise.all(
```
vaidotasp: This `await` looks to be doing nothing, we are calling `.then` on Promise.all anyway?

src/js/components/sharedComponents/UploadFile.tsx
```diff
@@ -89,13 +97,57 @@ const UploadFile = (): JSX.Element => {
...
+            const shapeFileFeatures: LayerFeatureResult = {
+              layerID: 'upload_file_features',
+              layerTitle: 'Upload File Features',
+              sublayerID: null,
+              sublayerTitle: null,
+              features: graphics
+            };
+
+            oldActiveFeatures.push(shapeFileFeatures);
+            dispatch(setActiveFeatures(oldActiveFeatures));
```
vaidotasp: This logic is leaving artifacts of previous upload. In the data panel if you upload a second file after the first, they both show up and it has some regressions, we need to clean out the activeFeature array with only single item (our shapeFileFeatures). Also from feature result we do not need anything else except attributes and geometry. Check out suggestion below.

```suggestion
            const graphics = mapController.generateGraphics(arcGISResults);

            const shapeFileFeatures: LayerFeatureResult = {
              layerID: 'upload_file_features',
              layerTitle: 'Upload File Features',
              sublayerID: null,
              sublayerTitle: null,
              features: graphics.map((g: __esri.Graphic) => {
                return { attributes: g.attributes, geometry: g.geometry };
              })
            };

            dispatch(setActiveFeatures([shapeFileFeatures]));
```

src/js/store/mapview/types.ts
```diff
@@ -86,6 +86,7 @@ export interface LayerFeatureResult {
   sublayerTitle: string | null;
   sublayerID: string | null;
   features: FeatureResult[];
+  geoStoreID?: string;
```
vaidotasp: May not need this anymore?
```suggestion

```

vaidotasp (changes requested):

src/js/helpers/MapGraphics.ts
```diff
@@ -32,16 +34,44 @@ export function setNewGraphic({
...
+  };
 
   allFeatures.forEach((feature: FeatureResult) => {
-    const symbol = (feature.geometry as any).rings
-      ? getCustomSymbol()
-      : polygonOrPointSymbol(feature.geometry.type);
+    const isPolygon =
+      (feature.geometry as any).rings || feature.geometry.type === 'polygon'
+        ? true
+        : false;
```
vaidotasp: not sure that we need `? true : false` that logic already evaluates to appropriate boolean.

src/js/helpers/MapGraphics.ts
```diff
@@ -32,16 +34,44 @@ export function setNewGraphic({
...
+    }
+  };
+
+  const setGeometry = (symbolType: string, geometry: __esri.Geometry): any => {
+    switch (symbolType) {
+      case 'polygon':
+        return new Polygon(geometry);
+      case 'point':
+        return new Point(geometry);
+    }
```
vaidotasp: if we have a switch statement, lets add default case too. it can try and return a Polygon?

vaidotasp (changes requested):

vaidotasp (approved):

PR #872, opened 2020-03-03
Type refactor for subscriptions

vaidotasp:
- Moving type definitions to already existing typings in redux store type file.
- Removing type casting for passed down props.
- Removed unneeded `key`


eyosef (approved):
LGTM

PR #876, opened 2020-03-05
Layer Loading Queries

vaidotasp:
- Reworked queries for active features on user click.  'feature', 'csv', 'geojson', 'scene' layers are queried with `layer.queryFeatures` method. ALL OTHER layers are queried with regular `QueryTask`
- due to the fact that we can have layer or sublayer being active at any given time, implemented logic accounting for it in terms of `LayerSelector` and overall `DataPanel`.
- documented perf and usability issues in separate ticket.

### DIFF
--- a/src/js/components/leftPanel/dataPanel/DataTabView.tsx
+++ b/src/js/components/leftPanel/dataPanel/DataTabView.tsx
@@ -26,12 +26,17 @@ const DataTabView = (props: DataTabProps): JSX.Element => {
   );
 
   const FeatureDataView = (): any => {
-    const activeLayer = activeFeatures[activeFeatureIndex[0]].layerID;
-    const activeLayerInfo = activeFeatures.find(f => f.layerID === activeLayer);
-    const activeLayerIndex = activeFeatures.findIndex(
-      f => f.layerID === activeLayer
-    );
+    const activeLayerInfo = activeFeatures[activeFeatureIndex[0]];
+    // const activeLayerInfo = activeFeatures.find(f => f.layerID === activeLayer);
 
+    function findLayer(f: any) {
+      if (f.sublayerID) {
+        return String(f.sublayerID) === String(activeLayerInfo.sublayerID);
+      } else {
+        return String(f.layerID) === String(activeLayerInfo.layerID);
+      }
+    }
+    const activeLayerIndex = activeFeatures.findIndex(findLayer);
     if (activeLayerInfo) {
       mapController.drawGraphic(
         activeFeatures[activeLayerIndex].features[activeFeatureIndex[1]]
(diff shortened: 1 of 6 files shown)


src/js/components/leftPanel/dataPanel/DataTabView.tsx
```diff
@@ -25,13 +26,18 @@ const DataTabView = (props: DataTabProps): JSX.Element => {
...
+    const activeLayerInfo = activeFeatures[activeFeatureIndex[0]];
 
+    //If layer has sublayers, we are using sublayerID to compare, otherwise it is layerID
+    function findLayer(f: LayerFeatureResult): boolean {
+      if (f.sublayerID) {
+        return String(f.sublayerID) === String(activeLayerInfo.sublayerID);
+      } else {
+        return String(f.layerID) === String(activeLayerInfo.layerID);
+      }
+    }
```
eyosef: I was thinking we could use a ternary operator?

```
const specificLayerID = f.sublayerID ? f.sublayerID : f.layerID

return String(f.specificLayerID) === String(activeLayerInfo.layerID);
```

src/js/components/leftPanel/dataPanel/DataTabView.tsx
```diff
@@ -95,11 +101,16 @@ const DataTabView = (props: DataTabProps): JSX.Element => {
...
-          f => f.layerID === layerID
-        );
+      function handleLayerSwitch(id: string): void {
+        //If layer has sublayers, we are using sublayerID to compare, otherwise it is layerID
+        function findLayer(f: LayerFeatureResult): boolean {
+          if (f.sublayerID) {
+            return String(f.sublayerID) === id;
+          } else {
+            return String(f.layerID) === id;
+          }
```
eyosef: I was thinking we could use a ternary operator?

```
const specificLayerID = f.sublayerID ? f.sublayerID : f.layerID

return String(specificLayerID) === id
```

src/js/helpers/DataPanel.ts
```diff
@@ -44,152 +17,150 @@ function esriQuery(url: string, queryParams: any): Promise<__esri.FeatureSet> {
...
+      if (
+        layer.type === 'feature' ||
+        layer.type === 'csv' ||
+        layer.type === 'geojson' ||
+        layer.type === 'scene'
+      ) {
+        try {
+          const featureResults = await layer.queryFeatures(queryParams);
+          //Ignore empty results
+          if (featureResults.features.length !== 0) {
```
eyosef: affirmative check? Could help us ID edge-cases later on (if `features` is undefined, or is an unexpected result, for example)
`if (featureResults.features.length > 0) {`

eyosef:
@vaidotasp what are your thoughts on using affirmative checks instead of `!==`? Just flagging as a general comment, feel free to let me know what you think!

vaidotasp:
> @vaidotasp what are your thoughts on using affirmative checks instead of `!==`? Just flagging as a general comment, feel free to let me know what you think!

I have no real preference, so we can do positive checks for various "let's see if we have any features or data" functions. But sometimes we need negative one `if layer.type !== 'graphics'` stuff like that.

eyosef (commented):
LGTM!

Issue #716, opened 2020-01-21
Left Panel - Layers Tab


SwampGuzzler:
- [X] Connect Layers Tab to Redux/Resources
- [X] Layer GROUP for each Group in Resources + the Layers from webmap
- [X] Checkbox per layer
- [X] Checkbox's `checked` based on associated layer's visibility (via Redux's `activeLayers`)
- [X] Style left panel (scaffolding) (#875 )
- [x] Style left panel (pixel-by-pixel detail)
- [x] Style left panel 'i' info content modal (pixel-by-pixel detail) #923 
- [x] Clicking on "i" icon shows a modal with additional information for the layer #922 
- [x] Clicking on "i" icon of `WEMAP_GROUP` shows a modal with additional information for the layer 
- [x] Layer controls display layer subtitle below its name #925 
- [x] Certain layers control canopy density > density slider implementation #884
- [x] Investigate layer density control list #886
- [x] Certain layers have start and end dates #929 
- [x] Certain layers have time period pickers (24H, 48H) #919 
- [x] Glad alert layer have unconfirmed alert toggle
- [x] Cover loss has dual slider

eyosef:
> * [x]  Clicking on "i" icon shows a modal with additional information for the layer

Updating this ticket to note that `WEBMAP_GROUP` is a type that doesn't have metadata to populate the `InfoContent` component. Since PROD has popup content for this layer group, we'll have to dig into this in a separate feature branch/PR to maintain SOC

Decision-tree @csphang provided to guide our approach;

**Webmap groups (for layers which come from webmap group, the following logic is used):**
- [ ] Full extended metadata from ArcGIS Online is used if the metadata is defined in ArcGIS Online for the item (https://doc.arcgis.com/en/arcgis-online/manage-data/metadata.htm)
- [x] If extended metadata is not defined, then the description and summary are used instead #932 
- [ ] If the layer added is an ArcGIS Feature Service or MapService and the service was added to the map via a URL, the metadata from the layer itself will be used.

**Non-Webmap groups (for layers which are configured through the configuration, the following logic is used):**
- [ ] If the Technical name for a layer is specified, then the metadata for the layer will be pulled from the GFW Metdata API
- [ ] If the layer is stored in the API, and includes a metadata URL the source used will be the metadata
- [x] If the technical name is not specified and the data is hosted through ArcGIS (Feature or MapService Layer) metadata from the layer will be used. Make sure your layer includes metadata before you publish it #922

[...]

PR #885, opened 2020-03-09
Canopy density control component

### LINKED ISSUE #884: Canopy density control component
Part of #716

- For  'TREE_COVER_LOSS', 'AG_BIOMASS', 'TREE_COVER' user can control canopy density which updates layers with the selected threshold.

image
image

vaidotasp:
Part of #716
Fixes #884

In Scope:
- Config out which layers have density control enabled
- `rc-slider` lib and associated types implementation
- density modal styling
- redux tracking of density value
- mapController function to update layers associated with density values


src/js/components/leftPanel/layersPanel/GenericLayerControl.tsx
```diff
@@ -13,6 +14,12 @@ const GenericLayerControl = (props: LayerControlProps): React.ReactElement => {
     (store: RootState) => store.mapviewState
   );
   const layer = allAvailableLayers.find(l => l.id === props.id);
+
+  //Determine if we need density control on this layer
+  let densityPicker = false;
+  if (layer) {
+    densityPicker = densityEnabledLayers.includes(layer.id);
```
eyosef: Could be a good fit?
``` javascript
  const densityPicker = layer && densityEnabledLayers.includes(layer.id);
```

src/js/controllers/mapController.ts
```diff
@@ -901,6 +901,20 @@ export class MapController {
     store.dispatch(renderModal(''));
   }
 
+  updateDensityValue(value: number): void {
+    densityEnabledLayers.forEach((layerId: string) => {
+      const layer: any = this._map?.findLayerById(layerId);
+      if (layer) {
+        console.log(layer);
```
eyosef: Should we remove the `console.log()` or is it a good fit to leave it in for debugging? Double-checking since there are some projects to log layer info for debugging purposes

configs/layer-config.ts
```diff
@@ -1 +1,8 @@
 export const allowedLayers = ['feature', 'dynamic', 'loss', 'gain']; //To be: tiled, webtiled, image, dynamic, feature, graphic, and custom (loss, gain, glad, etc)
+
+//Layer controls (IDS)
+export const densityEnabledLayers = [
+  'TREE_COVER_LOSS',
+  'AG_BIOMASS',
+  'TREE_COVER'
+];
```
eyosef, line 8: @vaidotasp Looks like the country config files might be tracking densityLayers via the property `GROUP_LCD`. Just flagging in case we need to revisit this later

vaidotasp: which property is tracking this in country config? I was looking through that but it was unclear to me which layers have this ability and which do not 🤔

src/js/components/leftPanel/layersPanel/GenericLayerControl.tsx
vaidotasp: 💯

configs/layer-config.ts
eyosef: Looks like `layerPanel[GROUP_LCD]` is tracking density (and I suspect a few other things). I noticed the function `updateDensity()` from `v1` leverages that object within the country config to maintain density, so I wanted to flag it for a future PR or ticket

```javascript
// LossControls.js, v1
  updateDensity (layer, density) {
    const { settings } = this.context;
    const layerGroups = settings.layerPanel;
    const layerConf = utils.getObject(layerGroups.GROUP_LCD.layers, 'id', this.props.layerId);
    let baseUrl = layerConf.url;
    baseUrl = baseUrl.split('tc')[0] + 'tc';
    baseUrl += density;
    baseUrl += '/{z}/{x}/{y}.png';    

    if (layer && layer.setUrl) {
      layer.setUrl(baseUrl);
    }
  }
```

eyosef (approved):
Looks great & runs great!

Issue #895, opened 2020-03-12
Attribute fetching for layers with no metadata


vaidotasp:
Q for @csphang.
There are examples of layers that do not have defined popup content, which prohibits us from efficiently filtering it and doing all required logic for language awareness.
Decided to do the following:
- We display ALL available attributes from such layer.

PR #896, opened 2020-03-12
WIP - Correct Attribute fetching

### LINKED ISSUE #888: Data tab - filter/format attributes
Data tab table should only show attributes that are allowed for layers in metadata object. Currently it shows all available attributes for any given feature.

- [x] remote data layers attribute filtering
- [x] webmap layer attribute filtering
- [ ] check for formatting options and format attribute data appropriately

vaidotasp:
Fixes #888

- Extracting popup information along with other metadata from the point of Layer creation. This information will be used for LegendConfig as well as Data Tab
- Types modifications to include expanded metadata objects
- Reworked DataPanel.ts, feature fetching logic for data panel:
1 . Made queries more sequential and less prone to duplicate.
2. Custom logic to fetch all available fields on the layer in order to extract OBJECTID. We will need object id when to fetch documents
3. Implemented logic where if user have no specified fields in the metadata, we return all fields. Decided to return all fields if there is no config. #895

### DIFF
--- a/src/js/controllers/mapController.ts
+++ b/src/js/controllers/mapController.ts
@@ -66,6 +66,8 @@ interface RemoteDataLayer {
     label: object;
     url: string;
     type: string;
+    popup?: object;
+    sublabel?: object;
     // [key: string]: object
   };
   dataLayer?: {
@@ -200,8 +202,16 @@ export class MapController {
                 let resourceGroup;
                 let url;
                 let type;
-
+                let metadata;
+                let popup;
+                let sublabel;
+                console.log(apiLayer);
                 if (apiLayer.dataLayer) {
+                  console.log('data layer is here');
+                  console.log(apiLayer);
+                  metadata = apiLayer.layer.metadata;
+                  popup = apiLayer.layer.popup;
+                  sublabel = apiLayer.layer.sublabel;
                   resourceId = apiLayer.dataLayer.id;
                   resourceTitle =
                     apiLayer.layer.label[appState.selectedLanguage];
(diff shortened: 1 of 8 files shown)


src/js/controllers/mapController.ts
```diff
@@ -200,8 +210,13 @@ export class MapController {
...
                 let url;
                 let type;
-
+                let metadata;
+                let popup;
+                let sublabel;
                 if (apiLayer.dataLayer) {
+                  metadata = apiLayer.layer.metadata;
+                  popup = apiLayer.layer.popup;
+                  sublabel = apiLayer.layer.sublabel;
```
vaidotasp, line 216: here is where we get the configs/metadata/sublabels etc etc etc

src/js/controllers/mapController.ts
```diff
@@ -131,6 +138,9 @@ export class MapController {
         () => {
           store.dispatch(isMapReady(true));
           //Set default language
+          const basemap = Basemap.fromId('osm');
+          //@ts-ignore
+          this._map?.basemap = basemap;
```
vaidotasp: ?

src/js/controllers/mapController.ts
```diff
@@ -562,14 +580,18 @@ export class MapController {
...
+      });
       //Replace all active features with our drawn feature, assigning custom layerID and Title
       const drawnFeatures: LayerFeatureResult = {
         layerID: 'user_features',
         layerTitle: 'User Features',
         sublayerID: null,
         sublayerTitle: null,
-        features: [event.graphic]
+        features: [event.graphic],
+        fieldNames: fieldNames
```
vaidotasp: do we need to return null for fieldNames?

vaidotasp:
sublayerID sublayerTitle, could those be optionals instead of union between null?

eyosef (commented):
@vaidotasp LGTM - let me know if you decide to implement the `interface` update, or the `util.ts` update for `DataPanel`. Then I'll do additional testing before approving and merging into `clean-slate`

eyosef (approved):
LGTM! Thanks for moving the helper functions into `helpers/dataPanel`!

Issue #793, opened 2020-02-14
ReferenceError: global is not defined


eyosef:
![Screen Shot 2020-02-14 at 2 40 08 PM](https://user-images.githubusercontent.com/9040867/74562039-e7dc6e00-4f37-11ea-9e32-84465646fcb2.png)

Do we need to configure `@arcgis/webpack-plugin` to account for `promiseUtils`?

- https://github.com/esri/arcgis-webpack-plugin#additional-features
- https://community.esri.com/thread/219307-web-worker-and-js-api-modules

vaidotasp:
Opened issue with the official arcgis plugin repo: 
https://github.com/Esri/arcgis-webpack-plugin/issues/80

Issue #772, opened 2020-02-10
Measure widget - coordinates(lat/long) widget


eyosef:
- `pointer-move` (not `onClick`) results should render in `modalCard.tsx`
- When user wants to convert `pointer-move` (not `onClick`) results to another unit, logic should handle this request and update the Redux store

PR #897, opened 2020-03-16
ESRI Coordinate widget implementation

### LINKED ISSUE #771: Measure widget - coordinates(lat/long) widget
- Utilize esri coordinate conversion and tracking widget for out of the box behavior.

image

vaidotasp:
Fix: #793
Fix: #772

- Changes to webpack config to resolve global module issues
- minor styling for measure widget
- attaching ESRI out of the box widget for coordinate tracking.

@eyosef please review and let me know if I missed anything in terms of code cleanup and general logic, you are much more familiar with this component that I am. Just want to make sure we have everything tidied up as we are using out of the box solution. Thanks!

### DIFF
--- a/src/js/controllers/mapController.ts
+++ b/src/js/controllers/mapController.ts
@@ -436,7 +436,7 @@ export class MapController {
   }
 
   attachCoordinatesWidget(domref: React.MutableRefObject<any>): void {
-    const ccWidget = new CoordinateConversion({
+    new CoordinateConversion({
       view: this._mapview,
       container: domref.current
     });
--- a/src/js/controllers/mapController.ts
+++ b/src/js/controllers/mapController.ts
@@ -7,6 +7,7 @@ import Graphic from 'esri/Graphic';
 import GraphicsLayer from 'esri/layers/GraphicsLayer';
 import SketchViewModel from 'esri/widgets/Sketch/SketchViewModel';
 import DistanceMeasurement2D from 'esri/widgets/DistanceMeasurement2D';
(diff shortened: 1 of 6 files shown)


eyosef:
```typescript
// measureContent.tsx
  switch (activeButton) {
    case 'area':
      selectedDropdown = areaUnitsOfLength;
      break;
    case 'distance':
      selectedDropdown = distanceUnitsOfLength;
      break;
    case 'coordinates':
      selectedDropdown = latitudeLongitudeUnits;
      break;
    default:
      selectedDropdown = defaultOption;
      break;
  }
```

@vaidotasp we'll want to update the case statement in `measureContent.tsx`  to set `selectedDropdown`  to `[]`  if the case is `coordinates` , since we're using the out of the box option

src/js/controllers/mapController.ts
```diff
@@ -434,6 +435,13 @@ export class MapController {
     }
   }
 
+  attachCoordinatesWidget(domref: React.MutableRefObject<any>): void {
+    const ccWidget = new CoordinateConversion({
```
eyosef:
```typescript
  attachCoordinatesWidget(domref: React.MutableRefObject<any>): void {
    new CoordinateConversion({
      view: this._mapview,
      container: domref.current
    });
  }

```
I'm thinking we won't need to set a new instance of `CoordinateConversion` to a constant, since it won't be accessed in other parts of `mapController.ts` or the app?

eyosef (approved):
Smooth fix - it could be worth doing a brown bag on @vaidotasp

PR #898, opened 2020-03-16
Ticket 870 - Document tab UI & logic 

### LINKED ISSUE #870: Document tab UI & logic
Screen Shot 2020-03-02 at 2 52 47 PM

- [X] On clicking the map, there is a query fired to determine if there are any documents associated with the specific provice/region. If there are, the document name will be listed in the document tab #898 
- [X] style scaffolding #898 
- [x] pixel-perfect styling
- [x] `flexbox` scaffolding (for mobile/tablet/desktop responsiveness)
- [x] `flexbox` scaffolding (for desktop responsiveness) #938

eyosef:
This PR fetches documents associated with a specific feature when the user selects the document tab in the left panel.

This PR fixes part of #870 

What this PR accomplishes;
- dynamically fetches documents associated with a specific feature
- populates results in the document tab view
- style scaffolding

What this PR doesn't accomplish
- styling is not pixel-perfect
- styling is not mobile/tablet responsive
- fetch logic may not account for the range of objectID names (`forets_production.objectid` vs `objectid` for example)


vaidotasp:
image

can we have KB size formatted like in PROD, I think we need to divide all values by `1000` to have it in KBs

src/js/interfaces/Attachment.ts
```diff
@@ -0,0 +1,24 @@
...
+export interface AttachmentWithURLProps {
+  id: number;
+  contentType: string;
+  size: number;
+  name: string;
+  iso: string;
+  layerTitle: string;
+  sublayerID: number;
+  specificFeatureID: number;
+}
```
vaidotasp: We can use intersection type here as AttachmentWithURLProps is a combination of the two. Thoughts?
```suggestion
export type AttachmentWithURLProps = Attachment & URLProperties;
```

src/js/store/appSettings/reducers.ts
```diff
@@ -36,7 +36,8 @@ const initialState: AppSettings = {
       widgetId: 'aa',
       uiParams: [{}]
     }
-  ]
+  ],
+  iso: 'CAF'
```
vaidotasp: does CAF mean specific country? can we have this as an empty string to be less confusing for the default state?

src/js/controllers/mapController.ts
```diff
@@ -986,6 +986,23 @@ export class MapController {
     }
   }
 
+  async getDocuments(
+    urlProperties: URLProperties
+  ): Promise<Array<Attachment> | null> {
+    const { iso, layerTitle, sublayerID, specificFeatureID } = urlProperties;
+    const endPoint = `https://gis.forest-atlas.org/server/rest/services/${iso}/${layerTitle}/MapServer/${sublayerID}/${specificFeatureID}/attachments?f=pjson`;
```
vaidotasp: What is the reason we are not using `queryAttachments()` built in ESRI method on FeatureLayers? https://developers.arcgis.com/javascript/latest/api-reference/esri-layers-FeatureLayer.html#queryAttachments
With this we would not need to keep this dynamic URL construction and just hit FeatureLayer endpoint and pass down object id that we care about? Can we check if that approach would return results?

vaidotasp (changes requested):

src/js/controllers/mapController.ts
```diff
@@ -1006,6 +1007,45 @@ export class MapController {
...
 
+  async getDocuments(
+    urlProperties: URLProperties
+  ): Promise<Array<Attachment> | null> {
+    const { sublayerID, specificFeatureID, layerID } = urlProperties;
+    let endPoint = '';
+    const layer = this._map?.findLayerById(layerID);
+
+    if ((layer as any).sublayers.length && specificFeatureID) {
+      const sublayer = (layer as any).sublayers.items.filter(
```
vaidotasp: let's move this function to some helper file instead of mapController. It is getting really big already.

We can just create a small function in mapController that accepts a layer ID and returns a layer it is found, but whole logic can live elsewhere.

src/js/components/leftPanel/DocumentsTabView.tsx
```diff
@@ -1,19 +1,113 @@
...
     (store: RootState) => store.appState.leftPanel
   );
-
+  const { activeFeatures, activeFeatureIndex } = useSelector(
+    (store: RootState) => store.mapviewState
+  );
   const tabViewIsVisible = tabViewVisible && activeTab === props.label;
-  return <>{tabViewIsVisible && <div>Documents Tab View</div>}</>;
+
+  const [featureCollectionIndex, featureIndex] = activeFeatureIndex;
```
vaidotasp:
```suggestion
  const [featureCollectionIndex] = activeFeatureIndex;
```
variable seems unused?

src/js/components/leftPanel/DocumentsTabView.tsx
```diff
@@ -1,19 +1,113 @@
...
+
+  const featureCollectionTitle = activeFeatures[featureCollectionIndex]
+    ? activeFeatures[featureCollectionIndex].sublayerTitle
+    : null;
```
vaidotasp: What if we do not have sublayerTitle? What if feature is on the "normal" layer? I think this needs to be this way:
"Layer Title: Sublayer Title", if no sublayer exists, we go with "Layer Title".
Thoughts?

vaidotasp (changes requested):

vaidotasp (approved):

PR #900, opened 2020-03-17
Ticket 892 - TimeSlider logic

### LINKED ISSUE #892: Layer Tab - TimeSlider component
- [x] using rc-slider component of dual marks - http://react-component.github.io/slider/examples/range.html #894 
- [x] default range is hardcoded from 2000 to 2018. #892 
- [x] updating dates seem to be modifying `options` property on BaseTileLayers and calling `refresh()` method on it (see example of updating canopy density in mapController) #900 
- [x] Play button and play functionality. Hitting play starts  going from lowest year to highest year, updating map. User can pause playing. #900 
- [x] UI - slider should have marks denoting the year (see below) #909 

image

eyosef:
This PR implements logic so the user can play/pause the time slider

Fixes part of #892 

What it accomplishes;
- Leverages a new approach to maintaining `setInterval()` via `useRef()`, without using [custom hooks](https://overreacted.io/making-setinterval-declarative-with-react-hooks/). I haven't found this approach yet online. This keeps the code clean and readable!
- Leverages local component state and the mapController to manage the time slider

What it does not accomplish;
- Styling changes

### DIFF
--- a/src/js/components/leftPanel/layersPanel/GenericLayerControl.tsx
+++ b/src/js/components/leftPanel/layersPanel/GenericLayerControl.tsx
@@ -28,7 +28,7 @@ const GenericLayerControl = (props: LayerControlProps): React.ReactElement => {
   const returnTimeSlider = (id: string): any => {
     switch (id) {
       case 'TREE_COVER_LOSS':
-        return <TimeSlider />;
+        return <TimeSlider layerID={id} />;
       default:
         return null;
     }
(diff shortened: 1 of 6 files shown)


vaidotasp (changes requested):
Couple issues I noticed:
1. This error in the console: 
image

2. In the PROD app, tile slider keeps going and going if user does not pause it. Meaning it keeps cycling over the range, our app stops at the last step.

3. Hitting pause resets the way the slider looks, but mapstate is not reflective of the 2000-2018? I do not see as many "pink" spots as I should.

vaidotasp (approved):
looks great

PR #907, opened 2020-03-18
Webmap sublayer split implementation

### LINKED ISSUE #906: Webmap layers and sublayer processing
We need to check if our webmap layers have sublayers and if that is the case, display those separately on the left panel. Giving user separate control over such sublayers in terms of visibility toggle, opacity slider, etc.

vaidotasp:
Fix #906

- in `mapController` when going over webmap layers we split sublayers out to add them individually to the `allAvailableLayers` array. This will allow us better control of such layers. Individual toggles, visibility sliders and so on.
- In order to make sure we have information that such layers were originally sublayers, we introduce two new properties on `allAvailableLayers`: `sublayer: boolean` and `parentID: string`. This will let us easier manipulate such layers later.
- Made adjustments where necessary in opacity, visibility functionality.


src/js/controllers/mapController.ts
```diff
@@ -156,23 +156,59 @@ export class MapController {
 
           const mapLayerObjects: LayerProps[] = [];
           this._map?.layers.forEach((layer: any) => {
```
eyosef: @vaidotasp what are your thoughts on moving the `.forEach()` into a designated helper function? I noticed the logic inside of this `.forEach()` is used on lines 158 and 435. Could be helpful if we need to update the logic later on?

vaidotasp: good idea. check the commit 8125afef323378a573f9ed78283991393920ed09

eyosef (approved):
Looks like this will make manipulating sublayers/layers much easier. Thanks @vaidotasp !

PR #908, opened 2020-03-18
Upload file types breaking deploy

vaidotasp:
- Removed sublayer related properties as we no longer need to set those to `null`
- added fieldNames to follow same way we create objects when we draw shape on the map


eyosef (approved):
LGTM :ok_hand:

PR #909, opened 2020-03-18
Ticket 892 - Time slider styling

### LINKED ISSUE #892: Layer Tab - TimeSlider component

eyosef:
This PR styles the time slider and renders the years when user clicks play

Fixes part of #892 

What it accomplishes;
- Leverages local component state to denote year when the user clicks play
- Resets marks to the default marks on pause
- Background is grey when user clicks grey

What it doesn't accomplish;
- Does use a SASS variable for the grey `background-color` because `#e9e9e9` is only used once

### DIFF
--- a/src/css/leftpanel.scss
+++ b/src/css/leftpanel.scss
@@ -191,6 +191,7 @@
     display: flex;
     padding: 0 1rem;
     align-items: center;
+    margin: 0rem 0rem 1rem 0rem;
 
     button {
       color: $base-yellow;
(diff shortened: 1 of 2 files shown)


vaidotasp:
looks and works well!

Issue #889, opened 2020-03-10
Implement Custom Legend


vaidotasp:
Reimplementation of ESRI Legend Widget to be able to use `legendConfig` that is pulled from every layer's metadata.
Ref: http://gis-gfw.wri.org/arcgis/rest/services/legends/MapServer/legend

- [x] Legend hides away if user clicks on the "eye" widget
- [x] User is able to collapse/expand legend by clicking on the title
- [x] Legend is scale aware (layers that are no in the right zoom level do not show on the legend)
- [x] Legend displays layers from webmap
- [x] Legend displays layers from GFW API
- [x] Legend displays layers from Services
- [x] Legend ignores layers that are MASK and legendLayers?
- [x] Legend displays all of the options from the config:
- [x] Legend works with feature server layers - #991
   - [x] - Basic
   - [x] - Line
   - [x] - Point
   - [x] - ServiceType
   - [x] - Cloropleth
   - [x] - Group
   - [x] - Gradient -> #921

- [x] legend is aware of "legendLayer": [0] - see wiki

vaidotasp:
@csphang I am not sure how to find layer that is using "GROUP" type of legend Config, maybe client or you can recall such usecase? or even point to API Layer that is using it so we can test that out.

vaidotasp:
https://production-api.globalforestwatch.org/v1/dataset/d808df62-96da-465f-a0e3-47b58aa6c5de/layer/bb5dc1f1-06b9-4117-943a-e2dd9341ea6b -- group test layer

PR #910, opened 2020-03-19
Custom legend implementation

vaidotasp:
Part of #889

In Scope
- Legend hides away if user clicks on the "eye" widget
 - User is able to collapse/expand legend by clicking on the title
 - Legend is scale aware (layers that are no in the right zoom level do not show on the legend)
 - Legend displays layers from webmap
 - Legend displays layers from GFW API
-  Legend displays all of the options from the config:
    - Basic
    - Line
    - Point
    - ServiceType

Out Of Scope
- Legend displays layers from Services
 - Legend is aware of "legendLayer": [0] - see wiki
 - Legend ignores layers that are MASK and legendLayers?
 Legend displays all of the options from the config:
    - Cloropleth
    - Group
    - Gradient


src/js/components/legend/generateLegendItems.tsx
```diff
@@ -0,0 +1,134 @@
...
+  }
+}
+
+const LegendItems = (props: LegendItemProps): JSX.Element => {
+  const { language } = props;
+  const items = props.visibleLayers.map(layer => {
+    if (layer.origin === 'webmap') {
+      const labelIcons = layer.legendInfo.map((item: any, i: number) => {
+        //deal with no label
+        item.label = item.label && item.label !== '' ? item.label : layer.title;
```
eyosef: I'm thinking we could check `.length` instead for an affirmative check?

```javascript
item.label = item.label && item.label.length ? item.label : layer.title;
```

vaidotasp: sounds good

eyosef (approved):
LGTM!

PR #918, opened 2020-03-24
Query feature optimization

vaidotasp:
This PR does not have corresponding issue but improves performance for query feature workflow.

- extracted function to check if layer is in scale into a helper to be reusable
- query for features is aware of the layer scale and does not query those layers that are visible but out of scale
- query for features is also aware of sublayer visibility, previously it was not.

### DIFF
--- a/src/js/controllers/mapController.ts
+++ b/src/js/controllers/mapController.ts
@@ -152,7 +152,12 @@ export class MapController {
             store.dispatch(setActiveFeatures([]));
             store.dispatch(setActiveFeatureIndex([0, 0]));
             store.dispatch(selectActiveTab('data'));
-            queryLayersForFeatures(this._mapview, this._map, event);
+            queryLayersForFeatures(
+              this._mapview,
+              this._map,
+              event,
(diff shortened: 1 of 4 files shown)


src/js/controllers/mapController.ts
```diff
@@ -152,7 +152,12 @@ export class MapController {
             store.dispatch(setActiveFeatures([]));
             store.dispatch(setActiveFeatureIndex([0, 0]));
             store.dispatch(selectActiveTab('data'));
-            queryLayersForFeatures(this._mapview, this._map, event);
+            queryLayersForFeatures(
```
vaidotasp: extract from mapview instead of sending separate

eyosef (commented):
LGTM! I did a performance test on Chrome and noticed the milliseconds spent on scripting decreased by 50% 🔥 

Once `queryLayersForFeatures()` is updated I'll merge this into `clean-slate`

eyosef (approved):
LGTM! Definitely saw a performance boost when comparing this branch to the alpha build!

PR #919, opened 2020-03-24
Ticket 915 - defined date range logic

### LINKED ISSUE #915: Reusable defined dates dropdown
Screen Shot 2020-03-20 at 3 05 23 PM

- [x] Create a reusable component with start/end dates #917 
- [x] Dynamically leverage across layers tab view #917 
- [x] Updating the dropdown updates the appropriate layer #919 
- [x] Toggling the feature layer is hooked up to sublayers #919 
- [x] Using the `layerVisibilitySlider` is hooked up to sublayers #919 
- [x] Updating the calendar updates the layer - also, `input type='date'` `min` is dynamically set #929 
- [x] Dynamically leverage in the analysis tab (workflow: user draws polygon, dropdown appears. Selecting certain options in the dropdown renders the start/end date component #952 
- [x] Styled to match comps #936

eyosef:
This PR implements logic for the dropdown in `DateRange.tsx`

This PR fixes part of #915 

What it accomplishes;
- Implements logic to support the custom date range feature
- Moved `DateRange.tsx` into `GenericLayerControl.tsx` for better control of rendering
- `LayerToggleVisibility.tsx` is now listening to `VIIRS` and `MODIS`
- `GenericLayerControl.tsx` resets `VIIRS` and `MODIS`
- `setCustomDateRange()` scaffolds logic for the calendar feature @vaidotasp let me know if I should remove it from this PR

What it does not accomplish;
- debugging the logic in `v1` that sets 72hr via the 7d endpoint - TBD!
- implementing styling of any kind - pure logic!

_Misc_ @vaidotasp feel free to let me know if I should move the logic in `mapController` to a file in the `helper` directory. I know we chatted about this already, but I included this reminder just in case, thanks so much!

### DIFF
--- a/src/js/components/leftPanel/layersPanel/DateRange.tsx
+++ b/src/js/components/leftPanel/layersPanel/DateRange.tsx
@@ -8,57 +8,69 @@ interface DateRangeProps {
   layer: LayerProps;
 }
 
+const returnDateToday = (): string => {
+  const dateToday = new Date().toLocaleDateString();
+  const [monthToday, dayToday, yearToday] = dateToday.split('/');
+  const dayTodayFormatted = dayToday.length === 1 ? `0${dayToday}` : dayToday;
+  const monthTodayFormatted =
+    monthToday.length === 1 ? `0${monthToday}` : monthToday;
+  const dateTodayFormatted = `${yearToday}-${monthTodayFormatted}-${dayTodayFormatted}`;
+
+  return dateTodayFormatted;
+};
+
 const DateRange = (props: DateRangeProps): JSX.Element => {
   const { layer } = props;
-
-  const returnDateToday = (): string => {
-    const dateToday = new Date().toLocaleDateString();
-    const [monthToday, dayToday, yearToday] = dateToday.split('/');
-    const dayTodayFormatted = dayToday.length === 1 ? `0${dayToday}` : dayToday;
-    const monthTodayFormatted =
-      monthToday.length === 1 ? `0${monthToday}` : monthToday;
-    const dateTodayFormatted = `${yearToday}-${monthTodayFormatted}-${dayTodayFormatted}`;
-
-    return dateTodayFormatted;
-  };
 
   const [startDate, setStartDate] = useState(returnDateToday());
   const [endDate, setEndDate] = useState(returnDateToday());
   const [renderCustomRange, setRenderCustomRange] = useState(false);
   const [definedRange, setDefinedRange] = useState('');
 
-  const setDate = (
-    updateStartDate: boolean,
-    e: ChangeEvent<HTMLInputElement>
-  ): void => {
-    if (updateStartDate) {
-      setStartDate(e.target.value);
-      mapController.setCustomDateRange(layer.id, e.target.value, endDate);
-    } else {
-      setEndDate(e.target.value);
-      mapController.setCustomDateRange(layer.id, startDate, e.target.value);
+  const updateStartDate = (e: ChangeEvent<HTMLInputElement>): void => {
+    setStartDate(e.target.value);
+    mapController.setCustomDateRange(layer.id, e.target.value, endDate);
+  };
+
+  const updateEndDate = (e: ChangeEvent<HTMLInputElement>): void => {
+    setEndDate(e.target.value);
+    mapController.setCustomDateRange(layer.id, e.target.value, endDate);
+  };
+
+  const resetSpecificDefinedRange = (): void => {
+    if (layer.id.includes('VIIRS')) {
+      mapController.resetVIRRSDefinedDateRange(layer.id);
+    }
+
+    if (layer.id.includes('MODIS')) {
+      mapController.resetMODISDefinedDateRange(layer.id);
     }
   };
 
   const setDefinedDateRange = (e: ChangeEvent<HTMLSelectElement>): void => {
-    // TODO [ ] fire query to correct URL
+    if (e.target.value === 'Select a date range') {
+      setDefinedRange('');
+      setRenderCustomRange(false);
+      resetSpecificDefinedRange();
+      return;
+    }
     setDefinedRange(e.target.value);
     setRenderCustomRange(false);
     mapController.setDefinedDateRange(layer.id, e.target.value);
   };
 
   /**
-   * TODO
-   * [ ] set min of input dynamically
+   * TODO [ ] set min of input dynamically
    */
 
   return (
     <>
       <div className="dropdown-wrapper">
-        <select onChange={(e): void => setDefinedDateRange(e)}>
-          <option selected={definedRange.length ? false : true}>
-            Select a date range
-          </option>
+        <select
+          onChange={(e): void => setDefinedDateRange(e)}
+          defaultValue={definedRange.length ? '' : 'Select a date range'}
+        >
+          <option value={'Select a date range'}>Select a date range</option>
           <option value={'24 hrs'}>Past 24 hours</option>
           <option value={'48 hrs'}>Past 48 hours</option>
           <option value={'72 hrs'}>Past 72 hours</option>
@@ -78,7 +90,7 @@ const DateRange = (props: DateRangeProps): JSX.Element => {
               value={startDate}
               min="2018-01-01"
               max={returnDateToday()}
-              onChange={(e): void => setDate(true, e)}
+              onChange={(e): void => updateStartDate(e)}
             />
           </div>
           <div className="date-section-wrapper">
@@ -89,7 +101,7 @@ const DateRange = (props: DateRangeProps): JSX.Element => {
               value={endDate}
               min="2018-01-01"
               max={returnDateToday()}
-              onChange={(e): void => setDate(false, e)}
+              onChange={(e): void => updateEndDate(e)}
             />
           </div>
         </>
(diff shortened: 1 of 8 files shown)


src/js/components/leftPanel/layersPanel/DateRange.tsx
```diff
@@ -13,44 +15,55 @@ const returnDateToday = (): string => {
...
+  const [definedRange, setDefinedRange] = useState('');
 
   const updateStartDate = (e: ChangeEvent<HTMLInputElement>): void => {
     setStartDate(e.target.value);
+    mapController.setCustomDateRange(layer.id, e.target.value, endDate);
   };
 
   const updateEndDate = (e: ChangeEvent<HTMLInputElement>): void => {
     setEndDate(e.target.value);
+    mapController.setCustomDateRange(layer.id, e.target.value, endDate);
```
eyosef:
```typescript
    mapController.setCustomDateRange(layer.id, startDate, e.target.value);
```

src/js/components/leftPanel/layersPanel/DateRange.tsx
```diff
@@ -13,44 +15,55 @@ const returnDateToday = (): string => {
...
   };
 
-  const setDefinedRange = (e: ChangeEvent<HTMLSelectElement>): void => {
-    console.log('setDefinedRange()', e.target.value);
-    // TODO [ ] fire query to correct URL
+  const setDefinedDateRange = (e: ChangeEvent<HTMLSelectElement>): void => {
+    if (e.target.value === 'Select a date range') {
```
eyosef: Don't need this option - default should be 24 hours

src/js/controllers/mapController.ts
```diff
@@ -258,6 +259,8 @@ export class MapController {
           });
 
           this.initializeAndSetSketch();
+          this.initializeAndSetVIIRSLayers();
+          this.initializeAndSetMODISLayers();
```
eyosef, line 281: Just noting that we've discussed conditionally firing these functions if the user's config files have `VIIRS` or `MODIS`. We've also discussed maintaining a single source of truth RE: available layers, instead of adding them directly to `this._map`

src/js/controllers/mapController.ts
```diff
@@ -1142,6 +1145,295 @@ export class MapController {
...
+    }
+
+    const sublayer24 = layer.sublayers.items.filter((sublayer: Sublayer) =>
+      sublayer.title.includes('24 hrs')
+    );
+
+    sublayer24.opacity = opacity;
+
+    if (layerID.includes('VIIRS')) {
+      const VIIRS48 = this._map?.findLayerById('VIIRS48');
```
vaidotasp: let's create an config object or array in layer configs helper that has listed all VIIRS fires layer IDS. That way we can use that same array, to iterate over when we create these layers or look for them by. Let me know if that does not make sense, happy to pair program.

src/js/controllers/mapController.ts
```diff
@@ -1142,6 +1145,295 @@ export class MapController {
...
+      const VIIRS72 = this._map?.findLayerById('VIIRS72');
+      const VIIRS7D = this._map?.findLayerById('VIIRS7D');
+
+      VIIRS48.opacity = opacity;
+      VIIRS72.opacity = opacity;
+      VIIRS7D.opacity = opacity;
+    } else if (layerID.includes('MODIS')) {
```
vaidotasp: Can we directly compare layerID with a string we expect instead of doing includes?

src/js/controllers/mapController.ts
```diff
@@ -1142,6 +1145,295 @@ export class MapController {
...
+  setDefinedDateRange(layerID: string, sublayerType: string): void {
+    if (!this._map) {
+      return;
+    }
+
+    const layer = (this._map.allLayers as any).items.filter(
+      (layer: LayerProps) => layer.id === layerID
+    )[0];
+
+    if (layerID.includes('MODIS')) {
```
vaidotasp: same here, compare layerID directly

src/js/controllers/mapController.ts
```diff
@@ -1142,6 +1145,295 @@ export class MapController {
...
+    // * leaving this in case we update the API
+    // layer.sublayers.items.forEach((sublayer: Sublayer) => {
+    //   // * Turns off all sublayers associated
+    //   // * with the selected layer (MODIS OR VIIRS)
+    //   if (!sublayer.title.includes(sublayerType)) {
+    //     sublayer.visible = false;
+    //   } else {
+    //     sublayer.visible = true;
+    //   }
+    // });
```
vaidotasp: I dont know if we need this, I doubt API will change and if it does by some chance in the future I think we can figure this out.

src/js/components/leftPanel/layersPanel/GenericLayerControl.tsx
```diff
@@ -27,6 +30,12 @@ const GenericLayerControl = (props: LayerControlProps): React.ReactElement => {
   //Determine if we need density control on this layer
   const densityPicker = layer && densityEnabledLayers.includes(layer.id);
 
+  useEffect(() => {
+    if (layer && (layer as any)?.visible === false) {
+      mapController.resetDefinedDateRange(layer.id);
+    }
+  }, [layer?.visible]);
```
vaidotasp: I do not think we need this useEffect that fires for any layer toggle action. We should have fires visibility being handled in mapController.toggleLayerVisibility or in the visibility component itself

vaidotasp (changes requested):
Couple suggestions in the review comments.

Also app crashes if I toggle any layer visibility on/off.

src/js/components/leftPanel/layersPanel/DateRange.tsx
```diff
@@ -56,7 +56,7 @@ const DateRange = (props: DateRangeProps): JSX.Element => {
           <option value={'24 hrs'}>Past 24 hours</option>
           <option value={'48 hrs'}>Past 48 hours</option>
           <option value={'72 hrs'}>Past 72 hours</option>
-          <option value={'7 days'}>Past week</option>
+          <option value={'7d'}>Past week</option>
```
vaidotasp: value set is `7d` but mapController seems to be expecting `7 days` can you check if that dropdown option ever fire appropriate functions?

src/js/components/leftPanel/layersPanel/LayerTransparencySlider.tsx
```diff
@@ -19,6 +19,13 @@ const LayerTransparencySlider = (
 
   const handleOpacityChange = (eventValue: any): void => {
     mapController.setLayerOpacity(layerID, eventValue, sublayer, parentID);
+
+    if (
+      layerID.includes('VIIRS_ACTIVE_FIRES') ||
+      layerID.includes('MODIS_ACTIVE_FIRES')
+    ) {
+      mapController.updateMODISorVIIRSOpacity(layerID, eventValue);
+    }
```
vaidotasp: this needs to be `if` `else`, right now even if we are dealing with fires layers, we are still kicking off setlayerOpacity above, that causes perf issues and errors.

src/js/controllers/mapController.ts
```diff
@@ -1183,6 +1188,289 @@ export class MapController {
...
+      return;
+    }
+
+    const sublayer24 = layer.sublayers.items.filter((sublayer: Sublayer) =>
+      sublayer.title.includes('24 hrs')
+    );
+
+    sublayer24.visible = false;
+
+    if (layer.id.includes('VIIRS_ACTIVE_FIRES')) {
```
vaidotasp: do we expect layer id to be something else here ? I am not sure why are we using `includes` and not a strict equality comparison

src/js/controllers/mapController.ts
```diff
@@ -1183,6 +1188,289 @@ export class MapController {
...
+            }
+          });
+        }
+        break;
+      default:
+        break;
+    }
+  }
+
+  resetDefinedDateRange(layerID: string): void {
```
vaidotasp: can we name this method something like `turnOffFireLayers`, I may be missing something but loooks like we are just turning off visibility but our method is saying it is resetting defined date range?

src/js/controllers/mapController.ts
```diff
@@ -1183,6 +1188,289 @@ export class MapController {
...
+          });
+        }
+        break;
+      case '7 days':
+        {
+          MODIS24.visible = false;
+          MODISLayerIDs.forEach(({ layerID }) => {
+            const specificLayer = this._map?.findLayerById(layerID);
+            if (specificLayer) {
+              if (specificLayer.id.includes('7d')) {
```
vaidotasp: can we do strict equality comparison again? I dont know if this can cause bugs, we are partially matching id name with two character string when we know that we need layer which ID is `MODIS7D` ?

vaidotasp (changes requested):
lookin' much better, find my comments below

src/js/controllers/mapController.ts
eyosef: @vaidotasp will do - we're turning off the visibility of the 24hr, 48hr, 72hr and 7day sublayer/feature layers for `VIIRS` or `MODIS`

vaidotasp (approved):

Issue #817, opened 2020-02-20
Add support for more layer types


eyosef:
`layer-config.js` confirms we don't support all layer types needed. `layerFactory.test.js` is a resource for confirming layer Factory works

version 1 `layerFactory.js` does support all layer types needed. It's a good resource for this ticket -> https://github.com/wri/gfw-mapbuilder/blob/master/src/js/utils/layerFactory.js

We need to support these types (list is not final and needs clarification):
- [x] map image
- [x] feature
- [x] imagery layers
- [x] loss layers
- [x] gain layers
- [x] tiled
- [x] vector??

Issue #763, opened 2020-02-10
Apply Layer Ordering


SwampGuzzler:
Layers added to the map respecting their GROUP and individual order

- [x] Layers ordered appropriately at initial load #920
- [ ] Layers ordered appropriately at language switch
- [ ] Layers ordered appropriately in report
- [ ] Layers ordered appropriately in share

SwampGuzzler:
@SwampGuzzler Let's **also** think about this after `Language-Toggle!`

PR #920, opened 2020-03-25
Biomass layer support + layer ordering logic

vaidotasp:
Partial fix for #763
Support new layer type that is tracked in #817

In scope:
- Added biodensity layer support (also should work with other ImageryLayers but needs testing)
- Added ability to control biodensity layer through canopy density slider
- Added lodash dependency for easier sorting
- Added layer ordering function and logic

Out of scope:
- Changing language flow does not respect layer ordering function for some reason.

### DIFF
--- a/src/js/components/mapWidgets/widgetContent/CanopyDensityContent.tsx
+++ b/src/js/components/mapWidgets/widgetContent/CanopyDensityContent.tsx
@@ -52,6 +52,7 @@ const CanopyDensityContent = (): JSX.Element => {
     dispatch(setCanopyDensity(value));
     //send % value to modify the layer
     mapController.updateDensityValue(markValueMap[value]);
+    mapController.updateBiodensityValue(value);
   }
 
   return (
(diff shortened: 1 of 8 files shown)


src/js/helpers/mapController/layerSorting.ts
```diff
@@ -0,0 +1,49 @@
+import Map from 'esri/Map';
+import { sortBy, flatten } from 'lodash-es';
```
eyosef: @vaidotasp looks like `@types/lodash-es` could help type `lodash-es`, otherwise we'll get the compile error `Could not find a declaration file for module 'lodash-es'`

src/js/components/mapWidgets/widgetContent/CanopyDensityContent.tsx
```diff
@@ -52,6 +52,7 @@ const CanopyDensityContent = (): JSX.Element => {
     dispatch(setCanopyDensity(value));
     //send % value to modify the layer
     mapController.updateDensityValue(markValueMap[value]);
+    mapController.updateBiodensityValue(value);
```
eyosef, line 55: When I test the UI, it looks like the biolayer isn't updating on the map in comparison to PROD

eyosef (approved):

PR #926, opened 2020-03-26
adding webtiled layers to layer factory

### LINKED ISSUE #817: Add support for more layer types

vaidotasp:
Partial Fix #817
This PR adds support for WebTileLayers. Testing with CMR config layer name `Couverture terrestre`

Layer loads okay and user is able to toggle it and change visibility. However seeing a lot of Tile errors in the console which may cause problems down below. Added as a discussion item with Christina. 
image

### DIFF
--- a/configs/layer-config.ts
+++ b/configs/layer-config.ts
@@ -1,4 +1,11 @@
-export const allowedLayers = ['feature', 'dynamic', 'loss', 'gain', 'image']; //To be: tiled, webtiled, image, dynamic, feature, graphic, and custom (loss, gain, glad, etc)
+export const allowedLayers = [
+  'feature',
+  'dynamic',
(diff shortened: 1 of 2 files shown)


eyosef (approved):
LGTM!

PR #734, opened 2020-01-29
Layers Panel - 716

vaidotasp:
#716 

- Created layer checkbox toggle component. Separate `tsx` file for component and `scss` file for its styling. It is a styled checkbox that gets its checked state from redux and on change calls mapController to toggle layers visibility.
- Created layer transparency component. On change calls mapController to modify layers opacity. Currently unstyled html range slider. We should consider `rc-slider` for easier controls, styling and features.
- imported infoIcon svg. Currently does not do anything, in the future it should fire modal with information about layer (see prod app for functionality)
- implementation of clear all, select all layers.
- 3 new redux actions: `clearAllLayers`, `updateVisibleLayers`, `allAvailableLayers`

Please check the type of change your PR introduces:
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

PR #928, opened 2020-03-27
Adding download data functionality to layer info panel

vaidotasp:
Partial for #734 

- Adding download data link for layer info panel


eyosef (approved):
Smart use of renaming a destructured property 🔥

PR #930, opened 2020-03-30
Share functionality implementation 

vaidotasp:
Partial fix for: #818

In scope
- Clicking on Share button, url is generated and these variables are currently tracked: 
   - Zoom, Lat , Long
   - Language
   - Active Tab
   - Visible layers and their opacities
   -  Canopy density

Upon map load all above variables with the exception of basemap are applied to the map creation process.
We currently do not track or apply the following variables:
- Basemap
- VIIRS, GLAD, MODIS, start/end dates

### DIFF
--- a/src/js/components/App.tsx
+++ b/src/js/components/App.tsx
@@ -10,7 +10,7 @@ import 'arcgis-js-api/themes/light/main.scss';
 import 'css/index.scss';
 import cameroon from '../../../configs/cameroon';
 import { overwriteSettings } from 'js/store/appSettings/actions';
-import { setLoggedIn } from 'js/store/appState/actions';
+import { setLoggedIn, setLanguage } from 'js/store/appState/actions';
 import { AppSettings } from 'js/store/appSettings/types';
 
 //TODO: SPinners should be SVGs in images/ folder that get imported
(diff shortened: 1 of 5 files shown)


src/js/components/App.tsx
```diff
@@ -36,9 +36,16 @@ const App = (props: AppSettings | any): JSX.Element => {
     //Determine which resources we are reading from
     //Read our local resources.js file And any external library resources (which are prioritized)
     dispatch(overwriteSettings({ ...cameroon, ...props }));
-    //Send that to our redux appSettings overwriting whatever is there
+    //Check URL for language param which comes in after user shares the application.
+    const langFromURL = new URL(window.location.href).searchParams.get('lang');
+    if (langFromURL) {
```
eyosef, line 41: @vaidotasp when I set the language and load the updated shareable URL, it looks the reloaded app doesn't update the language

vaidotasp: Thanks that should have been fixed with the latest commit! Re-test when you get a chance. Just heads up we still have issues with basemaps so english version may not load correctly, but at least it should listen properly to the URL hash.

Issue #863, opened 2020-02-28
Analysis Tab - Implementation


vaidotasp:
- [x] If user is logged in, we show subscribe button and launch the subscribe flow.
- [x] If user selects a feature or draws on the map, tab indicator shows that we have available 
- [x] User can click on PRINT REPORT buttons to launch its respective workflows.analysis => glowing SVG animation
- [x] If no feature is selected or nothing is drawn on the map, analysis tab shows "default" messaging
- [x] Tab shows all available analysis modules in a select dropdown (comes from resources)
- [x] Tab shows specific filters, controls for said analysis option (density, range etc) if available)
- [x] Upon clicking on analysis tab, geostore registration is done for a given geometry
- [x] Once it is registered, we `fetch` analysis data for vega widget from the api
- [x] Vega library accepts analysis data and displays a chart
- [x] Vega config is manipulated based on language, other custom PROD logic
- [x] Tab, content and charts are styled accordingly

PR #933, opened 2020-03-31
styling and UI elements for analysis tab

vaidotasp:
Part of #863
- Implements UI workflow and styling to get closer to PROD analysis tab

Out of scope:
- mobile styling
- required changes to vega chart api call
- ui elements rendering based on config (date range, density and datepicker)

### DIFF
--- a/src/js/components/leftPanel/analysisPanel/BaseAnalysis.tsx
+++ b/src/js/components/leftPanel/analysisPanel/BaseAnalysis.tsx
@@ -45,13 +45,9 @@ const BaseAnalysis = (): JSX.Element => {
     //On Base analysis tab we need to fire registration to geostore for the selected feature or the drawn/uploaded shape
     //Determine if we have the geostore already or we need to register it
     if (!activeLayer || activeFeature.attributes.hasOwnProperty('geostoreId')) {
-      console.log(
-        'we have no features or it already has geostoreID, do nothing'
-      );
       setGeostoreReady(true);
       return;
     } else {
-      console.log('feature does not exist, we need to register it');
       registerGeometry(activeFeature)
         .then(response => response.json())
         .then(res => {
@@ -66,7 +62,6 @@ const BaseAnalysis = (): JSX.Element => {
   }, [activeFeatures, activeFeatureIndex]);
 
   function runAnalysis() {
-    console.log('runAnalysis', selectedAnalysis);
     const mod = analysisModules.find(
       module => module.analysisId === selectedAnalysis
     );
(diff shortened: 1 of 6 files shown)


src/js/components/leftPanel/analysisPanel/BaseAnalysis.tsx
```diff
@@ -18,15 +20,18 @@ const AnalysisSpinner = (): React.ReactElement => (
 const BaseAnalysis = (): JSX.Element => {
   const dispatch = useDispatch();
   const [vegaSpec, setVegaSpec] = useState(null);
-  const [renderEditButton, setRenderEditButton] = useState(true);
+  const [renderEditButton, setRenderEditButton] = useState(false);
```
eyosef, line 23: @vaidotasp looks like this is referenced in `setActiveButton()` but `setActiveButton()` isn't being used. What are your thoughts on nixing this and `setActiveButton()`?

src/js/components/leftPanel/analysisPanel/BaseAnalysis.tsx
```diff
@@ -82,20 +87,76 @@ const BaseAnalysis = (): JSX.Element => {
...
-          </option>
-        );
-      })}
-    </select>
-  );
+  const AnalysisInstructions = (): JSX.Element | null => {
+    const currentAnalysis = analysisModules.find(
+      module => module.analysisId === selectedAnalysis
+    );
+    console.log(currentAnalysis);
```
eyosef: @vaidotasp there are a few `console.log()`, so I was thinking we could pair them down? If it's more beneficial to keep them, that's totally fine too. I've kept a few in previous PRs and found it to be very helpful

src/js/components/leftPanel/analysisPanel/analysisTranslations.ts
```diff
@@ -0,0 +1,46 @@
+export default {
+  RUN_ANALYSIS_BUTTON_TEXT: {
```
eyosef: @vaidotasp what are your thoughts on using camel case in this file? I noticed we've used camel case in previous content config files. Also, I was thinking we could name the object `runAnalysisButton`?

vaidotasp:
@eyosef made the requested changes, take a look when you get a chance! 
f48cadd

eyosef:
Looks like the `useEffect()` in `BaseAnalysis.tsx` might be the source of a memory leak. Workflow is to select a polygon on the map, and then select the analysis tab. Including a screenshot to support debugging;
Screen Shot 2020-04-02 at 2 27 35 PM

vaidotasp:
I dont think I touched useEffect on this PR. I propose creating a new issue to fix this bug.

eyosef:
Sounds good to me @vaidotasp

eyosef (approved):

PR #935, opened 2020-04-01
Ticket 931 - toggle bugfix

### LINKED ISSUE #931: Some layers toggle broke with recent merge
Some layers do not turn on anymore due to the error below.
Let's make sure that if we are toggling VIIRS/FIRES or any other "special handling case" layers we never hit `mapController.toggleLayerVisibility` helper.

image

eyosef:
This PR fixes errors we receive in the console when toggling a layer

Fixes part of #931 

What it accomplishes;
- Fixes toggle logic bug so layers are toggleable by conditionally firing `toggleVIIRSorMODIS()`

What it doesn't accomplish;
- Doesn't ID why there are certain layers that don't update on the map - I'm thinking this should go in a separate PR/ticket
    - Alertes SAD
    - Unités forestières d'exploitation
    - Assietes annuelles de coupe
    - Permis d'exploitation des carrières industrielles
    - Autorisation d'exploitation des carrières
    - Permis de recherches

### DIFF
--- a/src/js/components/leftPanel/layersPanel/LayerToggleSwitch.tsx
+++ b/src/js/components/leftPanel/layersPanel/LayerToggleSwitch.tsx
@@ -11,6 +11,12 @@ interface LayerToggleProps {
 
 const LayerToggleSwitch = (props: LayerToggleProps): React.ReactElement => {
   const { layerIsVisible, layerID, sublayer, parentID } = props;
+
+  const toggleVisibility = (): void => {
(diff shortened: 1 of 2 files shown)


vaidotasp:
Some of those layers that do not update on the map may be scale dependent! So if we are not in correct zoom level, we may be toggling those layers but nothing is happening. We may want to consider hiding such layers from left panel altogether, making left panel zoom scale aware, but something to think about in the different issue/PR definitely.

vaidotasp (approved):
🚀

PR #940, opened 2020-04-03
Layer Loading and language changing logic rework

vaidotasp:
The aim of this PR is to solve various inconsistency issues and duplication in types to do with layer creation and layer manipulation when user changes language.

There were a lot of inconsistencies and bugs introduced with the approach we had before that came apparent with the introduction of layers from resources.js file which have a direct URL (most of India's config does this).

Here is the high-level summary of work:
- Removed unneeded types and duplication of those, moving forward using `LayerProps` type for all layers we manipulate
- Reworked some of the layer creation login in mapController, added clarifying comments and made logic more sequential so it is easier to follow and understand what is being done step by step.
- Added support for service layers from config to to the legend
- Reworked a lot of `changeLang` method, previously it was buggy. Currently it works pretty well, and layer loading logic seems to be holding up okay (needs more testing).


configs/cameroon.js
```diff
@@ -670,6 +670,144 @@ module.exports = {
...
+      grouptype: 'default',
+      order: 1,
+      label: {
+        en: 'TEST',
+        fr: 'TEST',
+        es: 'TEST',
+        pt: 'TEST',
+        id: 'TEST',
+        zh: 'TEST',
+        ka: 'TEST'
```
eyosef: @vaidotasp I'm thinking we may need a clarifying comment that explains that `GROUP_TEST` was integrated to account for more layer types?

eyosef:
in `Documents.ts` we'll need to check the value of `attachmentInfos` to avoid logging errors in `.catch()`. Something like this;

```typescript
      const { attachmentInfos } = results;
      if (attachmentInfos) { // <- to ensure we're not mapping over undefined
        return attachmentInfos.map((attachment: Attachment) => {
          attachment.url = endPoint.replace(
            'attachments?f=pjson',
            `attachments/${results.attachmentInfos[0].id}`
          );
          return attachment;
        });
      }
```

configs/cameroon.js
vaidotasp: GROUP_TEST exists in development environment only, this cameroon.js file is not something that will be ever deployed. But I can put some comments if that's helpful!

eyosef: Yeah I was thinking comments could help since `GROUP_TEST` has layer info for India and the config file is largely for cameroon, but I'm flexible either way

eyosef:
@vaidotasp I noticed the `LayerTransparencySlider.tsx` isn't updating the legend with opacity for TEST layers

eyosef:
@vaidotasp Looks like `InfoContent.tsx` breaks when the user selects 'i'. Here is temporary logic we can use;

``` typescript
  const RenderLayerContent = (): JSX.Element | null => {
    if (layer.type === 'webmap') {
      return returnWebmapGroupContent();
    } else if (layer.type === 'dynamic') { // <- new
      // TODO update logic
      console.log('RenderLayerContent, dynamic layers!');
      return null;
    } else {
      return returnOtherGroupContent();
    }
  };
```

vaidotasp:
Document all bugs and I will address it within this PR! I do not want to introduce any regressions at this point, even if PR balloons out of scope.

vaidotasp:
> @vaidotasp Looks like `InfoContent.tsx` breaks when the user selects 'i'. Here is temporary logic we can use;
> 
> ```ts
>   const RenderLayerContent = (): JSX.Element | null => {
>     if (layer.type === 'webmap') {
>       return returnWebmapGroupContent();
>     } else if (layer.type === 'dynamic') { // <- new
>       // TODO update logic
>       console.log('RenderLayerContent, dynamic layers!');
>       return null;
>     } else {
>       return returnOtherGroupContent();
>     }
>   };
> ```

Which layers in particular?

eyosef:
> > @vaidotasp Looks like `InfoContent.tsx` breaks when the user selects 'i'. Here is temporary logic we can use;
> > ```ts
> >   const RenderLayerContent = (): JSX.Element | null => {
> >     if (layer.type === 'webmap') {
> >       return returnWebmapGroupContent();
> >     } else if (layer.type === 'dynamic') { // <- new
> >       // TODO update logic
> >       console.log('RenderLayerContent, dynamic layers!');
> >       return null;
> >     } else {
> >       return returnOtherGroupContent();
> >     }
> >   };
> > ```
> 
> Which layers in particular?

layers under 'TEST' category - all 3 break when 'i' is selected, because those particular layers doesn't have a `metadata` property

vaidotasp:
That makes sense, I will actually address that in the next PR. I had to reshuffle info content data fetching so those new layers start working properly. That will come with recent-imagery PR.

vaidotasp:
> in `Documents.ts` we'll need to check the value of `attachmentInfos` to avoid logging errors in `.catch()`. Something like this;
> 
> ```ts
>       const { attachmentInfos } = results;
>       if (attachmentInfos) { // <- to ensure we're not mapping over undefined
>         return attachmentInfos.map((attachment: Attachment) => {
>           attachment.url = endPoint.replace(
>             'attachments?f=pjson',
>             `attachments/${results.attachmentInfos[0].id}`
>           );
>           return attachment;
>         });
>       }
> ```

I am not clear why we need this change? I did not touch Documents file or anything that is relevant to it's functionality? I am probably missing something here

eyosef:
> > in `Documents.ts` we'll need to check the value of `attachmentInfos` to avoid logging errors in `.catch()`. Something like this;
> > ```ts
> >       const { attachmentInfos } = results;
> >       if (attachmentInfos) { // <- to ensure we're not mapping over undefined
> >         return attachmentInfos.map((attachment: Attachment) => {
> >           attachment.url = endPoint.replace(
> >             'attachments?f=pjson',
> >             `attachments/${results.attachmentInfos[0].id}`
> >           );
> >           return attachment;
> >         });
> >       }
> > ```
> 
> I am not clear why we need this change? I did not touch Documents file or anything that is relevant to it's functionality? I am probably missing something here

I was doing UI testing when I noticed when a user clicks India, and then the user clicks the document tab, we get an error stating `Cannot map over undefined`, so I figured an `if` statement could be a good fit to resolve the `.catch()` error here. I conducted general cross-app testing just in case. Let me know what you think!

vaidotasp:
@eyosef Merge conflicts fixed, checking for attachmentInfos, layer transparency should sync with legend. Not addressing the layer info panel because that got reworked on my recent-imagery PR which will follow this one.

eyosef (approved):

PR #948, opened 2020-04-10
Bugfix - Sketch widget

eyosef:
This PR fixes a bug that prevented the sketch widget from updating properly

What it accomplishes;
- leverages Redux to track whether or not the sketch widget is active
- updates logic where needed to ensure app doesn't break
- resets activeFeatures to `[]` when sketch widget is active

What it doesn't accomplish;
- doesn't determine why updating the store causes the sketch widget not to update properly


vaidotasp:
commit 276ca11 fixes incorrect Data Tab view behavior where component is re-rendered because redux state changed and tried to **draw** polygon on the map with a given geometry, however that geometry was coming from sketch widget which takes care of its own drawing. Thus it was drawing twice, code change should be self explanatory. Also added additional logic to register new geometry when user edits/saves drawn feature.

PR #949, opened 2020-04-10
Ticket 927 - Search quick UI

### LINKED ISSUE #927: Search widget needs quick UI
When the user selects a layer in the search widget dropdown, we need to render what `displayField` of the selected query to the user. This way the user knows what exactly is being queried

eyosef:
This PR renders what property the user can use to query a selected layer

PR fixes #927 

What it accomplishes;
- Search widget listens to user's layer selection and updates store
- `searchContent.tsx` listens to store and updates with UI content as needed

What it doesn't accomplish;
- translation logic (since this is a new, simple feature)

### DIFF
--- a/src/js/components/mapWidgets/widgetContent/searchContent.tsx
+++ b/src/js/components/mapWidgets/widgetContent/searchContent.tsx
@@ -18,9 +18,13 @@ const SearchContent: FunctionComponent = () => {
   const [formError, setFormError] = useState(false);
   const [latitudeInput, setLatitude] = useState('');
   const [longitudeInput, setLongitude] = useState('');
-  const selectedLanguage = useSelector(
-    (state: RootState) => state.appState.selectedLanguage
+  const { selectedLanguage, selectedSearchWidgetLayer } = useSelector(
+    (state: RootState) => state.appState
   );
+  const [selectedLayerInfo, setSelectedLayerInfo] = useState({
+    displayField: '',
+    layerTitle: ''
+  });
   const { title, buttonTitle, latitude, longitude } = searchContent[
     selectedLanguage
   ];
@@ -28,6 +32,10 @@ const SearchContent: FunctionComponent = () => {
   useEffect(() => {
     mapController.initializeSearchWidget(searchRef);
   }, []);
+
+  useEffect(() => {
+    setSelectedLayerInfo(selectedSearchWidgetLayer);
+  }, [selectedSearchWidgetLayer]);
 
   const setSearch = (): void => {
     if (latitudeInput === '' || latitudeInput === '') {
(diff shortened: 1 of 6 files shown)


src/js/components/mapWidgets/widgetContent/searchContent.tsx
```diff
@@ -18,9 +18,13 @@ const SearchContent: FunctionComponent = () => {
...
   const [longitudeInput, setLongitude] = useState('');
-  const selectedLanguage = useSelector(
-    (state: RootState) => state.appState.selectedLanguage
+  const { selectedLanguage, selectedSearchWidgetLayer } = useSelector(
+    (state: RootState) => state.appState
   );
+  const [selectedLayerInfo, setSelectedLayerInfo] = useState({
+    displayField: '',
+    layerTitle: ''
+  });
```
vaidotasp: why do we need local state here if we are already getting this info from redux store?

src/js/components/mapWidgets/widgetContent/searchContent.tsx
```diff
@@ -18,9 +18,13 @@ const SearchContent: FunctionComponent = () => {
   const [formError, setFormError] = useState(false);
   const [latitudeInput, setLatitude] = useState('');
   const [longitudeInput, setLongitude] = useState('');
-  const selectedLanguage = useSelector(
-    (state: RootState) => state.appState.selectedLanguage
+  const { selectedLanguage, selectedSearchWidgetLayer } = useSelector(
+    (state: RootState) => state.appState
```
vaidotasp: I have been going through the application and fixing our selectors. Basically this way that we have been extracting properties, will always cause a re-render of the component. We need to extract values separately, or use something like `reselect` lib.

```suggestion
  const selectedLanguage = useSelector(
    (state: RootState) => state.appState.selectedLanguage
  const selectedSearchWidgetLayer = useSelector(
    (state: RootState) => state.appState.selectedSearchWidgetLayer
```

src/js/components/mapWidgets/widgetContent/searchContent.tsx
```diff
@@ -29,6 +33,10 @@ const SearchContent: FunctionComponent = () => {
     mapController.initializeSearchWidget(searchRef);
   }, []);
 
+  useEffect(() => {
+    setSelectedLayerInfo(selectedSearchWidgetLayer);
+  }, [selectedSearchWidgetLayer]);
```
vaidotasp: If we read straight from the redux store, we may not need this use effect?

vaidotasp (changes requested):

src/js/components/mapWidgets/widgetContent/searchContent.tsx
eyosef: done and done

vaidotasp (approved):

PR #952, opened 2020-04-14
Analysis tab control - vega charts

### LINKED ISSUE #863: Analysis Tab - Implementation

vaidotasp:
Partial Fix #863 

- Add `reselect` library to utilize memoization for complicated Redux selectors
- Refactor selectors to prevent re-rendering in Header, Leftpanel, Analysis, Data Tab components
- Add uiParams for Analysis Tab: Date Picker, Year Slider, Canopy Density
- Add logic to generate unique URL based on uiParams that we use to fetch vega chart information

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,7 @@ module.exports = {
     'react-hooks/rules-of-hooks': 'error', // Checks rules of Hooks
     'react-hooks/exhaustive-deps': 'warn', // Checks effect dependencies
     '@typescript-eslint/ban-ts-ignore': 0,
-    '@typescript-eslint/no-var-requires': 0
+    '@typescript-eslint/no-var-requires': 0,
+    'react/prop-types': 0
   }
 };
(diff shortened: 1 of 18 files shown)


eyosef (approved):
LGTM

PR #972, opened 2020-04-21
Ticket 848 - edit upload

### LINKED ISSUE #848: Uploading a shapefile UI and logic

eyosef:
This PR implements logic to edit/delete uploaded files

Fixes part of #848 

What it accomplishes;
- When uploading a file with 1 graphic, user can edit/delete graphic
- When uploading a file with mutliple graphics, user can resize/delete graphic (sketch widget doesn't allow user's to edit multiple graphics)
- Small refactor to prevent multiple updates to the store

What it does not accomplish;
- as of now, `this._sketchVMGraphicsLayer` does not listen to `activeFeatures` property. I started working on this but realized it could be a heavy lift. I'm thinking I'll need to refactor a bit of `UploadFile.tsx` to account for this. I'll also need to refactor logic in `mapController.ts` to conditionally add/remove a selected graphic. This seems a bit involved so I'd like to suggest working on this on a separate feature branch. 
- 

_Just a heads up:_ @vaidotasp I added some dead code related to next steps. If you want me to continue working on added logic, I'll comment it in, but if not I'm happy to delete. Thanks!


vaidotasp (changes requested):
This below is caused by updating arcgis webpack plugin. Let's not do the update just yet while we are figuring these errors out!

image
image

eyosef:
> This below is caused by updating arcgis webpack plugin. Let's not do the update just yet while we are figuring these errors out!
> 
> image
> image

Just for future reference, I wanted to note that this was an issue tied to `arcgis-webpack-plugin` in `clean-slate`. @vaidotasp made a fix to `clean-slate` that was merged into this branch so we should be good to go. Thanks @vaidotasp !

vaidotasp (changes requested):
It is not compiling for me? Can you double check that we need that `multipleSelectionEnabled` on line 738 in mapController?

image

PR #977, opened 2020-04-22
Ticket 835 - template literal

### LINKED ISSUE #835: Template literals & SVGs in leftPanel.config.js
Screen Shot 2020-02-24 at 10 38 43 AM

Since we're leveraging JSX to render content in `leftPanel.config.js`, the draw icon SVG renders as plain text. mapbuilder version 1 uses the following approach, [which is now deprecated and could fail at any time](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href);

``` javascript
const drawIcon = '<svg class="svg-icon"><use xlink:href="#icon-draw-upload" /></svg>';
```

Is there a way to leverage template literals in `leftPanel.config.js`, or will we need added logic in `AnalysisTabView.tsx` to account for this?

eyosef:
This PR updates translations config to conditionally sandwich titles with the pen SVG.

Fixes #835 

What it accomplishes;
- conditionally sandwiches titles with the pen SVG. Since this is config'd translations we can always expect the logic to work

### DIFF
--- a/configs/leftPanel.translations.ts
+++ b/configs/leftPanel.translations.ts
@@ -1,5 +1,3 @@
-const drawTool = 'DRAW_ICON_HERE';
-
 export const analysisContent = {
   en: {
     analyzeExistingShapeTitle: 'Analyze a shape on the map',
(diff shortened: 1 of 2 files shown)


vaidotasp:
🚀

PR #998, opened 2020-04-27
Ticket 984 - Analysis and Doc Tabs "Pulsing" effect

### LINKED ISSUE #966: Analysis and Doc Tabs "Pulsing" effect
When user clicks on a feature, if we have analysis and doc tabs available, it should give a 'pulsing' effect and display an svg on those tabs. Like in PROD:

image

### LINKED ISSUE #984: Narrative Tab styling improvements
Our version:
image

Prod version:
image

eyosef:
This ticket implements the blinking yellow dot in the analysis and docs tab.

Fixes #966 

What it accomplishes;
- Uses local component state & css to conditionally apply the pulsing tab icon.


vaidotasp (changes requested):
There is some padding up top missing on the svgs based on what we have in PROD.

image
prod:
image

src/js/components/leftPanel/LeftPanel.tsx
```diff
@@ -81,20 +111,60 @@ interface TabsProps {
...
+
+  const activeFeatureIndex = useSelector(
+    (store: RootState) => store.mapviewState.activeFeatureIndex
+  );
+
+  const [featureCollectionIndex, featureIndex] = activeFeatureIndex;
+
+  useEffect(() => {
+    const specificFeature =
+      activeFeatures[featureCollectionIndex]?.features[featureIndex];
```
vaidotasp: no sure why we need to look for that active feature here? Can we simply check if `activeFeatures` length is larger than 0?

vaidotasp (changes requested):
There are a lot of re-renderings happening here which may out of the scope of this PR. But one thing I noticed is if I select a feature. Then quickly hit analysis tab and then hit documents tab, analysis tab starts pulsing again after a split second once geostore id gets registered and it rerenders.

src/js/components/leftPanel/LeftPanel.tsx
eyosef: I know we chatted about this yesterday, but I wanted to document this for future reference - this logic helps track when the user selects 'next' or 'previous' in the data tab. This way the doc/analysis tab can render the 'pulsing' effect with each selection

vaidotasp (changes requested):
Application opens up Document tab by default? Is that intended behavior?

Also I am getting this error on load?
image

vaidotasp:
> There is some padding up top missing on the svgs based on what we have in PROD.
> 
> image
> prod:
> image

Has this been addressed ?

eyosef:
@vaidotasp I've run this branch locally and can confirm the padding has been addressed. Below is a screenshot of the feature branch;

Screen Shot 2020-04-30 at 11 08 49 AM

below is a screenshot of the yoga site
Screen Shot 2020-04-30 at 11 10 27 AM

vaidotasp (approved):
I removed the timeout and it seems to be functioning okay, we still have a lot more complexity I think than we need to but this may be something for the future to improve. Merging in 👍

PR #1017, opened 2020-05-05
MODIS and VIIRS Fires alerts improvements

vaidotasp:
Partial for #818

This PR does the following:
1. Refactors how we create and interact with all MODIS and FIIRES layers. Previously we created them as FeatureLayers by directly accessing URL within MapImageServer. That introduced performance issues and very non-optimal queries, 1Year layer did not work at all. With this change all layers are created as MapImageLayers with their consecutive sublayer ids.
2. Redux tracking for date ranges related to these layer groups
3. Users can share date range for both layers with the usual share flow.

### DIFF
--- a/src/js/controllers/mapController.ts
+++ b/src/js/controllers/mapController.ts
@@ -1471,9 +1471,6 @@ export class MapController {
   }
 
   initializeAndSetVIIRSLayers(): any {
-    //TODO: check for state here and apply def expression
-    // const { viirsStart, viirsEnd } = store.getState().appState.leftPanel;
-    // const defExpression = `ACQ_DATE > date '${viirsStart}' AND ACQ_DATE < date '${viirsEnd}'`;
     const viirsLayers = VIIRSLayerIDs.map(({ id, url, layerIds }) => {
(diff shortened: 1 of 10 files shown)


src/js/controllers/mapController.ts
```diff
@@ -1433,96 +1437,102 @@ export class MapController {
...
+    const MODIS24 = this._map!.findLayerById('MODIS_ACTIVE_FIRES');
 
     switch (sublayerType) {
       case '24 hrs':
         {
           MODIS24.visible = true;
-          MODISLayerIDs.forEach(({ layerID }) => {
-            const specificLayer = this._map?.findLayerById(layerID);
+          MODISLayerIDs.forEach(({ id }) => {
+            const fireLayer = this._map?.findLayerById(id);
```
eyosef: It's sort of nitpicky but I was thinking `fireLayer` could be renamed to `modisLayer`

vaidotasp: ya I like it, I like super specific names so when we read the code we know exactly what is going on

PR #1027, opened 2020-05-07
Ticket 934 - Layers Info content & logic

### LINKED ISSUE #934: Layers Info Content and Logic
Description of the logic flow:
**Webmap groups** (for layers which come from webmap group, the following logic is used):

1. Full extended metadata from ArcGIS Online is used if the metadata is defined in ArcGIS Online for the item (https://doc.arcgis.com/en/arcgis-online/manage-data/metadata.htm)
2. If extended metadata is not defined, then the description and summary are used instead
3. If the layer added is an ArcGIS Feature Service or MapService and the service was added to the map via a URL, the metadata from the layer itself will be used.

**Non-Webmap groups** (for layers which are configured through the configuration, the following logic is used):
1. If the Technical name for a layer is specified, then the metadata for the layer will be pulled from the GFW Metdata API
2. If the layer is stored in the API, and includes a metadata URL the source used will be the metadata
3. If the technical name is not specified and the data is hosted through ArcGIS (Feature or MapService Layer) metadata from the layer will be used. Make sure your layer includes metadata before you publish it

Checklist:
- [x] Webmap - Checks for ArcGIS Online metadata. This is on hold until we talk to Christina as PROD apps fail to do this fetch we need to make sure we actually need this function
- [x] Webmap - Description and summary is used using endpoint for each layer `layer.url/?f=pjson`
- [x] Webmap - not sure how No3 is different from No.2 it warrants a conversation, for now this item is on hold
- [x] Non-Webmap - if layer has technical name, hit GFW Metadata API to retrieve info
- [x] Non-Webmap - if layer does not have technical name, use metadata that comes from GFW API along with layer info
- [x] Non-Webmap - if layer has no technical name, and was created through URL in the config, use metadata on layer like in webmap No.2 scenario

eyosef:
This ticket implements the logic flow for ticket 934 - fixes #934 

What it accomplishes;
- implements logic flow for `InfoContent.tsx`
- works when testing w/appID for Christina's test app & `clean-slate`'s original webmap ID
- If logic workflow doesn't work, content returns 'No information available'

What it doesn't accomplish;
- While I did test this logic across multiple webmap IDs, since there are many ways to create maps, it's possible we may run into an edge case some day. Just flagging for future reference.


src/js/controllers/mapController.ts
```diff
@@ -120,6 +120,7 @@ export class MapController {
     this._map = new WebMap({
       portalItem: {
         id: webmapID
+        // id: 'a81c9f897c7041d59f1076687d2bde16' // * Christina's test app ID
```
vaidotasp: let's remove this line
```suggestion
```

src/js/helpers/mapController/miscLayerHelpers.ts
```diff
@@ -100,35 +100,83 @@ export async function extractWebmapLayerObjects(
...
+          maxScale,
+          minScale,
+          sublayer: false,
+          portalItemID: layer.portalItem.id
+        });
+      } else {
+        legendInfo = layer.layerId
+          ? legendInfo.layers.find((l: any) => l.layerId === layer.layerId)
+              .legend
+          : legendInfo;
```
vaidotasp: can you pull in latest `clean-slate` and remove the changes from this file? I think I have a solution for this bug already in place!

src/js/components/sharedComponents/InfoContent.tsx
```diff
@@ -1,33 +1,175 @@
...
+
+interface ContentError {
+  error: string;
+}
+
+type Content = MetadataContent | ContentError;
+
+const getMetadata = async (layer: any): Promise<any> => {
+  let content: any;
+  const metadataURL = `https://www.arcgis.com/sharing/rest/content/items/${layer.portalItemID}/info/metadata/metadata.xml`;
```
vaidotasp: the domain "https://www.arcgis.com" should be coming from the redux `appSettings.sharinghost` can you double check that it is the case in PROD as well?

src/js/components/sharedComponents/InfoContent.tsx
```diff
@@ -1,33 +1,175 @@
...
+      (element: any) => element.name === 'idPurp'
+    ).elements[0].text;
+
+    content = {
+      title,
+      functionOrPurpose,
+      overview
+    };
+  }
+  if (!metadataExists && layerMetadataExists) {
```
vaidotasp: let's do `else if (layerMetadataExists) {`

src/js/components/sharedComponents/InfoContent.tsx
```diff
@@ -1,33 +1,175 @@
...
+    const title = element.elements
+      .find((element: any) => element.name === 'idCitation')
+      .elements.find((subElement: any) => subElement.name === 'resTitle')
+      .elements[0].text;
+    const overview = element.elements.find(
+      (element: any) => element.name === 'idAbs'
+    ).elements[0].text;
+    const functionOrPurpose = element.elements.find(
+      (element: any) => element.name === 'idPurp'
+    ).elements[0].text;
```
vaidotasp: we can extract this into a function or something like that because we are using this exact code multiple times?

src/js/components/sharedComponents/InfoContent.tsx
```diff
@@ -1,33 +1,175 @@
...
+        functionOrPurpose,
+        overview
+      };
+    }
+  } else {
+    // * if portalItemID exists, fetch info w/ it
+
+    if (layer.portalItemID) {
+      const results = await fetch(
+        `https://www.arcgis.com/sharing/rest/content/items/${layer.portalItemID}/?f=pjson`
```
vaidotasp: same comment about sharinghost from redux applies here.

vaidotasp (changes requested):

src/js/components/sharedComponents/InfoContent.tsx
eyosef: @vaidotasp confirmed - that's what's happening in `clean-slate`. I'll implement this now

src/js/components/sharedComponents/InfoContent.tsx
```diff
@@ -1,33 +1,136 @@
...
+  tags?: string;
+  overview?: string;
+}
+
+interface ContentError {
+  error: string;
+}
+
+type Content = MetadataContent | ContentError;
+type SharingHost = string | undefined;
```
vaidotasp: instead of declaring this here, can we do the following: 1) in `appSettings/types` lets change sharinghost from optional parameter to required 2) can we add default sharinghost in appSettings/reducers to be 'https://www.arcgis.com'. This way we can read it directly from appSettings and we do not need to declare type or potentially have it as undefined because it will always be defined in that default reducer.

src/css/leftpanel.scss
```diff
@@ -533,12 +533,17 @@
 .layers-control-container {
   color: $dark-grey;
 
+  .buffer {
+    padding-bottom: 1rem;
+  }
+
```
vaidotasp: instead of creating this buffer div that adds padding can we remove it and add `margin-top: 1rem` to the `.layers-control-checkbox` class?

vaidotasp (changes requested):

vaidotasp (approved):

PR #1030, opened 2020-05-08
Ticket 1024 - Subscribe to alerts flow styling

### LINKED ISSUE #1024: Subscribe to alerts flow styling
Match/improve upon subscribe to alerts styling. PS. let's not forget about customColorTheme in this modal!

eyosef:
This PR styles the subscribe to alerts feature. Fixes #1024 

What it accomplishes;
- matches comps
- improved UI/UX - back/forward buttons, form validation

What it doesn't accomplish;
- pixel perfection (but it's close) - I noticed this feature leans on `display: block` on `PROD` so I implemented flex grid instead. Styled to align with comps closely

### DIFF
--- a/src/css/alertCarousel.scss
+++ b/src/css/alertCarousel.scss
@@ -34,8 +34,7 @@
     ul {
       display: grid;
       list-style: none;
-      grid-template-columns: 8rem 8rem;
-
+      grid-template-columns: 1.5fr 1.5fr;
       li {
         font-size: 0.875rem;
         font-weight: 400;
@@ -89,7 +88,6 @@
           appearance: none;
           padding: 0.75rem;
           padding-right: 2rem;
-          background-color: transparent;
           cursor: pointer;
           height: 2rem;
           width: auto;
--- a/src/css/alertCarousel.scss
+++ b/src/css/alertCarousel.scss
@@ -1,0 +1,5 @@
+@import './variables.scss';
+
+.alert-carousel-container {
+  display: flex;
+}
--- a/src/css/alertCarousel.scss
+++ b/src/css/alertCarousel.scss
@@ -2,4 +2,119 @@
 
 .alert-carousel-container {
   display: flex;
+
+  .orange-button {
+    &.custom {
+      width: 3rem;
+      font-weight: 400;
+    }
+
+    &.inactive {
+      opacity: 0.5;
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


src/css/alertCarousel.scss
```diff
@@ -0,0 +1,123 @@
...
+
+    p {
+      font-weight: 400;
+      line-height: 1.7;
+    }
+
+    ul {
+      display: grid;
+      list-style: none;
+      grid-template-columns: 8rem 8rem;
```
vaidotasp: you actually can achieve the "place content in two equally sized columns that take up all space" like this `grid-template-columns: 1fr 1fr`

src/js/components/leftPanel/dataPanel/subscribeToAlerts/NameYourSubscription.tsx
```diff
@@ -36,35 +36,42 @@ const NameYourSubscription = (
...
+          </select>
+        </div>
+        <div className="buttons-wrapper">
+          <button
+            onClick={(): void => setPrevStep()}
+            className="orange-button custom esri-icon-left-arrow"
+          />
+          <button
+            onClick={(): void => setNextStep()}
+            disabled={subscriptionName.length ? false : true}
```
vaidotasp: do we need that ternary? `subscription.length should evaluate to boolean already? or am I missing some edge case?

vaidotasp (changes requested):
In addition to those two small things, let's make sure that subscription flow respects `customColorTheme` from the config. You can check out various other components that are listening to this for example `PrintReportButton.tsx`

src/js/components/leftPanel/dataPanel/subscribeToAlerts/NameYourSubscription.tsx
eyosef: `subscriptionName.length` would return `true` and disable the button when we need the button to only be disabled when `subscriptionName.length` is 0. I can refactor so it looks like this;

```typescript
          <button
            onClick={(): void => setNextStep()}
            disabled={subscriptionName.length === 0}
```

vaidotasp (changes requested):
can we make all buttons/icons in the flow respect customColorTheme? that seems to be in line with what is in prod right now.
image
image
image

vaidotasp (approved):

PR #1044, opened 2020-05-11
Ticket 1029 - google analytics integration

### LINKED ISSUE #1029: Add google analytics to the application
Seems like GA tag is in the config, let's check prod how GA is implemented and what is tracked and re-implement it.

eyosef:
This PR integrates google analytics and ensures the uniqueID is maintained by config. 
Fixes part of #1029 

What it accomplishes;
- default `analyticsCode` is the code used in PROD
- default is overwritten dynamically by config

What it doesn't accomplish;
- doesn't yet send events to a GA dashboard (coming soon!)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -22,9 +22,6 @@
         'https://www.google-analytics.com/analytics.js',
         'ga'
       );
-
-      ga('create', 'UA-62288390-15', 'auto');
-      ga('send', 'pageview');
     </script>
     <!-- End Google Analytics -->
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -1,6 +1,32 @@
...
+      })(
+        window,
+        document,
+        'script',
+        'https://www.google-analytics.com/analytics.js',
+        'ga'
+      );
+
+      ga('create', 'UA-62288390-15', 'auto');
+      ga('send', 'pageview');
```
vaidotasp: doesnt this cause google analytics event to be fired twice? once when index.html executes and once in App.tsx? should we just leave App.tsx one?

vaidotasp (changes requested):

src/index.html
eyosef: That's an excellent point - I'll remove this from `index.html`

vaidotasp (approved):

Issue #963, opened 2020-04-20
Vega chart library dependency


vaidotasp:
Currently, due to issues with ARCGIS plugin/webpack config, we are loading vega lib from a CDN. Ideally we would want to find a solution that is perhaps more stable, bundling chart lib in the bundle?

PR #1045, opened 2020-05-11
Swap vega link to CDN link to be ready for PROD

vaidotasp:
Swapping vega chart link to the one that is CDN. Based on their docs, that is what we should be using.

Partially related and addressing this: #963


eyosef (approved):
LGTM!

PR #1054, opened 2020-05-14
Ticket 880 - draw message

### LINKED ISSUE #880: Drawing a polygon - can we set a custom message?
Screen Shot 2020-03-06 at 3 07 33 PM

workflow: draw a polygon. Select 'edit' to edit the polygon. Custom message appears

eyosef:
This PR implements a workaround to ticket #880 

What it accomplishes;
- When a user's mouse hovers over their drawn polygon, and the polygon is in edit mode, a popup will appear with the lat/long info
- styled to match comps

What it doesn't accomplish;
- does not match PROD logic (couldn't figure out how to get the popup to appear only when the mouse hovers over a point in the drawn polygon)

### DIFF
--- a/src/js/components/mapWidgets/userPointPopup.tsx
+++ b/src/js/components/mapWidgets/userPointPopup.tsx
@@ -1,0 +1,59 @@
+import React, { useEffect, useState } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+
+import { setRenderPopup } from 'js/store/appState/actions';
+
+import { RootState } from 'js/store';
+
+const UserPointPopup = (): JSX.Element | null => {
+  const dispatch = useDispatch();
+  const [coordinates, setCoordinates] = useState<__esri.Point | null>(null);
+
+  const userCoordinates = useSelector(
+    (state: RootState) => state.mapviewState.userCoordinates
+  );
+  const renderPopup = useSelector(
+    (state: RootState) => state.appState.renderPopup
+  );
+
+  useEffect(() => {
+    if (userCoordinates) {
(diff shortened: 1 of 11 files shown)


src/js/components/mapWidgets/userPointPopup.tsx
```diff
@@ -0,0 +1,59 @@
...
+  );
+  const renderPopup = useSelector(
+    (state: RootState) => state.appState.renderPopup
+  );
+
+  useEffect(() => {
+    if (renderPopup && userCoordinates) {
+      const latitude = (userCoordinates as any).latitude
+        .toString()
+        .match(/^-?\d+(?:\.\d{0,2})?/)[0];
```
vaidotasp: what does this regex do?

vaidotasp (changes requested):
image
- let's get rid of scrollbars? User would not be able to see all the information

- Also position is not tracked if user moves a vertex and hovers over it.

src/js/components/mapWidgets/userPointPopup.tsx
eyosef: This returns the last 2 decimals without rounding up or down. For example, `14.123456` becomes `14.12`

vaidotasp: I would rather increase precision to 5 decimals than use regex that we have to document somewhere. `const latitude = userCoordinates.latitude.toFixed(5)` ?

eyosef: Excellent point - consider it done!

vaidotasp (approved):

PR #1056, opened 2020-05-15
Ticket 954 - Translations audit

### LINKED ISSUE #954: Language translation audits
- [x] Go over all application and find missing translations #1056 
- [x] Config out all translations in one big folder/file for easier editing as we go through this exercise

eyosef:
This PR implements translations where missing. Fixes #954 

What it accomplishes;
- share widget, coordinates DMS/DD, canopy density picker and info widget listen to `selectedLanguage`

### DIFF
--- a/src/js/components/mapWidgets/widgetContent/coordinatesDDSection.tsx
+++ b/src/js/components/mapWidgets/widgetContent/coordinatesDDSection.tsx
@@ -1,10 +1,22 @@
 import React from 'react';
+import { useSelector } from 'react-redux';
+
+import { RootState } from 'js/store/index';
 
 import { DDSectionProps } from 'js/interfaces/coordinateForm';
+
+import { coordinatesContent } from 'configs/modal.config';
 
 import { ReactComponent as TrashCanIcon } from 'images/trashCanIcon.svg';
 
 export default function DDSection(props: DDSectionProps): JSX.Element {
+  const selectedLanguage = useSelector(
+    (state: RootState) => state.appState.selectedLanguage
+  );
+  const { latitudeLabel, longitudeLabel } = coordinatesContent[
+    selectedLanguage
+  ];
+
   const {
     ddSection,
     degreeSymbol,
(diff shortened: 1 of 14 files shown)


src/js/components/sharedComponents/CanopyDensityPicker.tsx
```diff
@@ -27,12 +32,29 @@ const CanopyDensityPicker = (props: CanopyDensityPickerProps): JSX.Element => {
   return (
     <>
       <div>
-        {props.label && <span>Change canopy density</span>}
-        <button
-          className="canopy-density-picker"
-          style={{ backgroundColor: `${customColorTheme}` }}
-          onClick={handleDensityButtonClick}
-        >{`> ${markValueMap[density]}%`}</button>
+        {props.label &&
```
vaidotasp: This component is too complex. Why are we checking for props.label existence? I see from the types that this type is required, do we have a scenario that props.label does not exist?

Also this is just two spans and a button. I do not believe we need this map function here with index checking and so on.
```
<span>{displayLabel[0]}</span>
<button></button>
<span>{displayLabel[1]}</span>
```
would that work?

src/js/components/mapWidgets/widgetContent/coordinatesDDSection.tsx
```diff
@@ -1,10 +1,22 @@
...
+import { RootState } from 'js/store/index';
 
 import { DDSectionProps } from 'js/interfaces/coordinateForm';
 
+import { coordinatesContent } from 'configs/modal.config';
+
 import { ReactComponent as TrashCanIcon } from 'images/trashCanIcon.svg';
 
 export default function DDSection(props: DDSectionProps): JSX.Element {
+  const selectedLanguage = useSelector(
```
vaidotasp, line 13: we can get selected Language info from parent component and just pass down language as prop, reducing number of `useSelectors` is a good goal have if that is reasonable.

vaidotasp (changes requested):
couple small things and this is good to go.

src/js/components/sharedComponents/CanopyDensityPicker.tsx
eyosef: yep - I'll implement that now. I noticed `props.label` was used to conditionally render static text, so I included it just in case,  but I'll also remove that logic and update the props interface, etc for `CanopyDensityPicker`

PR #1057, opened 2020-05-15
Refactor mouse location tracking when user edits drawn/uploaded shape on

vaidotasp:
the map, simplification of event tracking and popup rendering


eyosef (approved):
LGTM!

PR #1074, opened 2020-05-28
Ticket 1070 - mobile header styling

### LINKED ISSUE #1070: Mobile - Header
I think we should remove all elements from the header with the exception of LOGO. Even though in the old version we have those elements. They are very squished together and the same functionality is achieved through "MORE" button.

Prod:
image
Rewrite:
image

eyosef:
On mobile, app hides elements in header except for the title, subtitle & icon. Fixes #1070

### DIFF
--- a/src/css/header.scss
+++ b/src/css/header.scss
@@ -247,6 +247,18 @@
   }
 }
 
+@media screen and (max-width: $tablet-device) {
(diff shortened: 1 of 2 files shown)


vaidotasp (approved):
looks good!

PR #1079, opened 2020-06-01
Ticket 1076 - legend button & mobile styling

### LINKED ISSUE #1076: Mobile - Legend needs to be mobile friendly too
Here is how it looks in PROD (CMS):
Screen Shot 2020-05-30 at 11 41 02 AM
Screen Shot 2020-05-30 at 11 40 47 AM

Essentially it adds an additional icon to the widget list that toggles legend, which slides out from the bottom.

eyosef:
This PR ensures the legend widget is togglable for mobile workflow. Fixes #1076 

What it accomplishes;
- listens to user's window dimensions - conditionally updates the store based on window dimensions width
- legend widget button renders on mobile only

What it doesn't accomplish;
- doesn't ID whether `hideLegend` should continue to be in the `appSettings` reducer


vaidotasp (changes requested):
`hideLegend` setting in the config is used for when user wants to have legend minimized at the start of the application.

There are some bugs involved with that: 
1 - Set `hideLegend: true` in the config
2 - Load app in Mobile view
3 - Upon clicking Legend icon, legend content is not visible.

vaidotasp (changes requested):
if you have `hideLegend: true`, that removes legend altogether from the desktop view. it is suppose to just toggle it down and minimize it not completely remove it?

I feel like it would be just easier to use so other redux value if redux is needed here to track when we should be displaying legend mobile view or legend desktop view? Whatever is the most simple way of achieving this without breaking the existing functionality.

configs/resources.js
```diff
@@ -17,6 +17,7 @@ export default {
   mapThemeIds: '', // e.g. 1c38ba1095fe49e3ba234bf9105c1077;c76d788b7487476bae4d09a4e933be19
   mapThemes: '', // e.g. Forest Atlas of Cameroon;Forest Atlas of Equatorial Guinea
   narrative: '',
+  hideLegend: true,
```
vaidotasp: Let's have default to be the FALSE value here.

src/js/components/legend/Legend.tsx
```diff
@@ -50,6 +59,24 @@ const Legend = (): JSX.Element => {
...
+
+    window.addEventListener('resize', handleResize);
+    window.removeEventListener('resize', handleResize);
+  }, []);
+
+  useEffect(() => {
+    const { width } = windowDimensions;
+
+    if (width < 475 && hideLegend) {
+      setLegendOpen(hideLegend);
```
vaidotasp, line 82: Can you add a small comment here of why we are doing this hide Legend checks for the future? I feel like people who will be working on this code at some point may get confused. Couple lines would suffice.

src/js/components/legend/Legend.tsx
```diff
@@ -50,6 +59,24 @@ const Legend = (): JSX.Element => {
...
     []
   );
+
+  useEffect(() => {
+    const handleResize = () => {
+      setWindowDimensions(getWindowDimensions());
+    };
+
+    window.addEventListener('resize', handleResize);
+    window.removeEventListener('resize', handleResize);
```
vaidotasp: This snippet adds event listener and removes it immediately? I am assuming we want the event listener for the duration of the component lifecycle? If so, we need to return the clean up function per React Docs: https://reactjs.org/docs/hooks-effect.html

```suggestion
    () => window.removeEventListener('resize', handleResize);
```

vaidotasp (changes requested):
Looks great, just a couple tiny changes and we can merge this in. Good work!

vaidotasp (approved):

Issue #1083, opened 2020-06-04
Make the layer tab the default even when the info window is activated


csphang:

vaidotasp:
This should be solved in normal app loading and share app functionality, it is suppose to go to layer. Can we re-test with the latest link?

PR #1109, opened 2020-06-10
Remove active tab sharing function, it will always default to layers

vaidotasp:
Remove active tab sharing function, it will always default to layers


vaidotasp:
#1083

PR #1180, opened 2021-03-09
Planet basemap implementation

vaidotasp:
Pending "proxy" solution. This is a WIP for the front-end, will be adjusted as needed once proxy is in.


vaidotasp:
#1181

PR #1274, opened 2022-05-04
removed the modis layer and updated wordpress config

davidlstarr:


davidlstarr:
Need to completely remove all of the modis methods and config files after session 1 knowledge transfer

PR #1335, opened 2023-02-15
refactor: logic for getting report attributes

anthonyamaro15:
Jira ticket: https://gfw.atlassian.net/browse/MAPB-130
test link: https://alpha.blueraster.io/gfw-mapbuilder/pull-requests/MAPB-130/?appid=0174a49d49244b1cb57f7491c173144c

@davidlstarr  once this is approved I will remove test web map before merging,

Test cases tested:
1. Drawing polygon
2. Selecting polygon

### DIFF
--- a/configs/urls-config.ts
+++ b/configs/urls-config.ts
@@ -1,3 +1,0 @@
-export const URLS = {
-  GEOSTORE: 'https://production-api.globalforestwatch.org/v1/geostore/',
-};
--- a/configs/urls-config.ts
+++ b/configs/urls-config.ts
@@ -1,0 +1,3 @@
+export const URLS = {
+  GEOSTORE: 'https://production-api.globalforestwatch.org/v1/geostore/',
+};
--- a/configs/resources.js
+++ b/configs/resources.js
@@ -1,8 +1,5 @@
 export default {
-  // webmap: 'de85e3fcc07948238aa6c1afd2a4ceb0',
-  webmap: 'fcd1ee44193c447ba639b3a910a0ef4d',
-  // webmap: '8e83c8ef36624556859cb605a093337b',
-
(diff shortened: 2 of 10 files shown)


davidlstarr:
@anthonyamaro15 I'm seeing just an outline for the polygon on the report page. Is that correct?

Screen Shot 2023-02-17 at 10 30 06 AM

configs/urls-config.ts
```diff
@@ -0,0 +1,3 @@
+export const URLS = {
+  GEOSTORE: 'https://production-api.globalforestwatch.org/v1/geostore/',
```
davidlstarr: Is this file really needed if its only used once in the code?

davidlstarr (changes requested):

anthonyamaro15:
@davidlstarr  for the outline polygon, we can create a new ticket for that since the issue is not related to the print issue.

davidlstarr:
@anthonyamaro15 Is the client aware of this issue?

configs/urls-config.ts
anthonyamaro15: @davidlstarr I will be adding endpoints here as I go to isolate endpoints to a single file

anthonyamaro15:
@davidlstarr  I don't think so, this only happens to some of the polygons

configs/urls-config.ts
davidlstarr: Are there any other endpoints in the app atm? I honestly don't think this is needed and we can create this when there are multiple endpoints.

davidlstarr:
@anthonyamaro15 Go ahead and create a Jira ticket in the backlog for this item

davidlstarr (approved):

PR #1338, opened 2023-02-22
feat: setup wms support for webmap and config files

davidlstarr:
Jira ticket: https://gfw.atlassian.net/browse/MAPB-77
Config file Test link: https://alpha.blueraster.io/gfw-mapbuilder/pull-requests/MAPB-77
Webmap Test link: https://alpha.blueraster.io/gfw-mapbuilder/pull-requests/MAPB-77/?appid=8e83c8ef36624556859cb605a093337b


davidlstarr:
@anthonyamaro15 need to revert resource.js file before merging

src/js/controllers/mapController.ts
```diff
@@ -896,7 +906,7 @@ export class MapController {
       layer = this._map
         ?.findLayerById(parentID)
         //@ts-ignore -- sublayers exist
-        ?.allSublayers.items.find((sub: any) => sub.id === layerID);
+        ?.allSublayers.items.find((sub: any) => sub.id == layerID);
```
anthonyamaro15: is there a reason why you removed `===` to `==` ? 
The difference between these two is that the abstract equality operator doesn’t care about the data types, whereas the strict equality operator does.

src/js/components/legend/Legend.tsx
```diff
@@ -79,39 +63,26 @@ const Legend = (): JSX.Element => {
     //IDS of layers you want to specifically ignore in the legend
     const ignoredLayers = ['RECENT_IMAGERY'];
     const visibleLayers = allAvailableLayers
-      .filter(l => l.visible)
-      .filter(l => !ignoredLayers.includes(l.id))
-      .filter(l => layerIsInScale(l, scale));
+      .filter((l) => l.visible)
+      .filter((l) => !ignoredLayers.includes(l.id))
+      .filter((l) => layerIsInScale(l, scale));
+    console.log('visibleLayers', visibleLayers);
```
anthonyamaro15: could we remove console?

davidlstarr: yes

src/js/controllers/mapController.ts
davidlstarr: I removed because the `layerID` was coming across as a string but was looking for a number to match with the `sub.id`.

anthonyamaro15: @davidlstarr  yes because if you compare 
```js
 '1' == 1  output true // which in reality these two values are not the same
```

you can just do
```js
Number(sub.id) === Number(layerID)
```

davidlstarr: cool I will make the change and test

anthonyamaro15 (approved):

davidlstarr:
NEED TO RESET resource.js file BEFORE MERGING!!!

PR #1396, opened 2024-05-01
fix: 1391 deps fix

anthonyamaro15:
test link: https://alpha.blueraster.io/gfw-mapbuilder/pull-requests/1391-deps-fix


davidlstarr:
We had to apply `--legacy-peer-deps` because Webpack will need to update Webpack which could cause a lot of work to resolve the issues.

Screenshot 2024-05-01 at 11 07 24 AM

PR #1434, opened 2025-09-29
feat: layer configuration readme

anthonyamaro15:
PR covers:
- Documentation for adding layer configuration on updated workflow


eKerney:
This looks good overall.  I've followed along with directions and it the parameters seemed to match well with the documentation.  Perhaps we should move all docs except README into the docs dir to clean things up.

anthonyamaro15:
> This looks good overall. I've followed along with directions and it the parameters seemed to match well with the documentation. Perhaps we should move all docs except README into the docs dir to clean things up.

Good idea, I'll create a separate PR for moving all documentation into it

PR #1471, opened 2026-06-03
update: security vulnerabilities

anthonyamaro15:
PR covers:

Remove unused deps (cypress, uuid, compression-webpack-plugin), upgrade copy-webpack-plugin 6→14 and html-loader 0.5.5→5.1.0, and add a uuid override for the sockjs path. No runtime code changed; build, typecheck, and dev-server render all verified.
Screenshot 2026-06-03 at 1 16 02 PM
Screenshot 2026-06-03 at 1 17 05 PM


ziveo (approved):
👍

PR #1485, opened 2026-08-24
Fix/tree cover loss canopy density

anthonyamaro15:
Changing the canopy density on Tree Cover Loss had no effect — the map always rendered 30% data.

The tile API stopped honoring the legacy `tcd_{value}` path after v1.11. It redirects to the dynamic endpoint as `?implementation=tcd_{value}`, which v1.12+ ignores. Not introduced by the 2025 update; that only bumped v1.12 → v1.13.

Verified against the tile server — `tcd_75` is byte-identical to the unfiltered default, the query param is not:

| request | md5 |
|---|---|
| `v1.13/tcd_75/4/2/4.png` | `1fabd5df…` (= default) |
| `v1.13/dynamic/4/2/4.png` | `1fabd5df…` |
| `v1.13/dynamic/4/2/4.png?tree_cover_density_threshold=10` | `484700b2…` |
| `v1.13/dynamic/4/2/4.png?tree_cover_density_threshold=75` | `8abe484c…` |

## Fix

Point the layer at the dynamic endpoint with `?tree_cover_density_threshold={thresh}`, reusing the `{thresh}` placeholder `TreeCoverLossLayer.getTileUrl` already substitutes. The density now lives on the layer instead of being baked into `layerConfig.url`, so toggling the layer off and on keeps the selection.

Second, unrelated fix in the same layer: `filter()` read `this.view.zoom`, which is fractional mid-animation, while `intensityBank` is keyed by integer zoom. The lookup missed and left pixels uncolored. It now colors by the tile's own level.

## Testing

Manually verified — loss pixels visibly thin out at 75% and fill in at 10%, and Network shows `dynamic/…?tree_cover_density_threshold=` with no `tcd_` and no redirect hop.
Test link: https://d828dk8ho9zgd.cloudfront.net/gfw-mapbuilder/pull-requests/fix/tree-cover-loss-canopy-density/index.html


ziveo (approved):
👍

ziveo (approved):
👍

